PR #10, opened 2020-06-29
Fixed bugs for blobstore and added fields in blog

smarquez-jpg:


tayyaba17 (approved):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -103,13 +104,19 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
 
       // Get the URL of the image that the user uploaded to Blobstore.
       String imageUrl = getUploadedFileUrl(request, "image");
 
       //Get replies.
-      ArrayList<String> messageReplies = getParameter(request, "replies", "").split(",");
-
+      String messageRepliesString = getParameter(request, "replies", "");
+	    String messageRepliesArray[] = messageRepliesString.split(",");
```
Andrewmob2703-zz, line 116: I may be wrong but I think using split assumes that all replies are obtained in one round, and you're trying to separate them? Since some replies can come later, I think using params to get and add them to the ArrayList probably will be better

Andrewmob2703-zz (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
smarquez-jpg: I'm not sure if I understand, but the once the form is submitted all the data becomes static, so if we were to have a dynamic list of responses that updated regularly, we would need to implement some other functionality for that

Andrewmob2703-zz: Oh okay, makes sense

PR #12, opened 2020-06-30
Authentication

tayyaba17:
Trying to figure out why nothing happens when I click the button to onload showPostAuthentication()


tayyaba17:
For the showPostForm() function in the script. The code inside the if statement still gets executed even when isLoggedIn is false

portfolio/src/main/webapp/script.js
```diff
@@ -79,4 +79,19 @@ function fetchBlobstoreUrlAndShowForm() {
         messageForm.action = imageUploadUrl;
         messageForm.classList.remove('hidden');
       });
-}
\ No newline at end of file
+}
+
+function showPostForm() {
```
smarquez-jpg, line 84: I think the reason that you still see the blogcontent is because there's nothing saying that it should be hidden when loading the webpage

tayyaba17: in the CSS the display for this container is set to none. When the page initially loads, I don't see anything until i click the button, which should set the display to block (viewable) only if the user is logged in, but currently it changes it regardless

smarquez-jpg: Then, is the problem that isLoggedIn just always returns true?

peizhaoou: Can we try to print out the "isLoggedIn" variable and see its value?

peizhaoou: I suspect this will print out just a text here, not necessarily the bool you want.

tayyaba17:
The HomeServlet.file will be deleted.

smarquez-jpg (approved):

portfolio/src/main/webapp/script.js
```diff
@@ -82,16 +82,11 @@ function fetchBlobstoreUrlAndShowForm() {
 }
 
 function showPostForm() {
-  fetch('/login').then(response => response.text()).then((isLoggedIn) => {
+  fetch('/login_status').then(response => response.json()).then((isLoggedIn) => {
     if (isLoggedIn) {
       console.log("inside IF")
```
peizhaoou, line 87: Nit: if this is only used during your debugging, it may be good to just remove them when you try check in code.

peizhaoou: We usually only keep the critical debugging print statements in the codebase.

tayyaba17: My bad! I deleted them after commiting

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -25,7 +25,6 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       String urlToRedirectToAfterUserLogsOut = "/";
       String logoutUrl = userService.createLogoutURL(urlToRedirectToAfterUserLogsOut);
 
-      response.getWriter().println(isLoggedIn);
       response.getWriter().println("<p>Hello " + userEmail + "!</p>");
       response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
     } else {
```
peizhaoou, line 30: Is the "urlToRedirectToAfterUserLogsIn" variable used just once below? If so, worth removing the variable altogether. If you deem that as a constant that will be used elsewhere, make it a class member variable.

portfolio/src/main/java/com/google/sps/servlets/LoginStatusServlet.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/login_status")
```
peizhaoou, line 12: What is the difference between the LoginStatusServlet & LoginServlet?

tayyaba17: LoginServlet is the page where the user will login and is html based. LoginStatus is json based and simply returns the whether or not the user is logged in

peizhaoou (approved):

PR #13, opened 2020-06-30
Not done yet but have basic logic mapped out for tests

smarquez-jpg:


tayyaba17 (commented):
I think most of these changes are my old file changes to add the tabs. I think the new stuff was mostlyt around line 204, right? If so, LGTM.

tayyaba17:
Please lmk if I missed stuff. It says some files not shown so I am not sure what that means. Don't forget to add our hosts!

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class HashTableTest {
+  @Test
+  public void noComments() {
```
tayyaba17, line 31: I think something we all need to touch base on is when to use comments vs. posts, just so we are clear what the function is actually doing. Seems to me that this should be noPosts()?

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+    Hashtable actual = new Hashtable<>();
+    int expected = 0;
+
+    Assert.assertEquals(expected, actual.size());
+  }
+
+  @Test
+  public void oneBlogPostForEachTag() {
+    //Return a Hashtable with LinkedList of size one for each tag.
+    Hashtable<String, List<BlogMessage>> actual = new Hashtable<>();
```
tayyaba17: LinkedList

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+    actual("Classical") = BlogMessage(3, "Classical", "Classical music makes you smarter", "", "Andrew", ArrayList<String>(), System.currentTimeMillis());
+
+    Set<String> keys = actual.keySet();
+    for(String key : keys){
+      Assert.assertEquals(1, actual.get(key).size);
+    }
+  }
+
+  @Test
+  public void multipleBlogPostsForFewTags() {
```
tayyaba17, line 54: Seems like this test is just checking size and not order too, I think you can remove that comment since you're testing order in the next test

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+    List<String> expected = new ArrayList<>();
+    for(int i = 0; i < 4; i++){
+      long int time = System.currentTimeMillis();
+      actual("Pop") = BlogMessage(i, "Pop", "Test Message", "", "Steven", ArrayList<String>(), time);
+      expected.add(BlogMessage(i, "Pop", "Test Message", "", "Steven", ArrayList<String>(), time));
+    }
+    
+    Set<String> keys = actual.keySet();
+    for(int i = 0; i < expected.size(); i++){
+      Assert.assertEquals(expected(i), actual.get("Pop").message);
```
tayyaba17: Should you be adding the actual message to expected instead of the whole BlogMessage? It seems like you are asserting each index of expected to be equal to the message in pop but each index in expected is actually a BlogMessage

martiKallas (changes requested):
There a lot of files and changes here that are unrelated to your tests. You should add a .gitignore file to the repository. Andrew might already be doing this, so coordinate with him. .gitignore files prevent compilation artifacts and secret files from being checked in to the public repository. This will remove all the portfolio/target/... files. See the top directory of your starter project for a .gitignore file that will work for you.

PR #16, opened 2020-07-02
Implement LinkedHashMap to hold BlogMessages under tags

Andrewmob2703-zz:


portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public final class BlogHashMap {
+  private static final String[] internalTags = {"#general", "#wellbeing", "#music", "#games"};
```
peizhaoou, line 27: If internalTags is supposed to be a constant-like variables (which I think is your assumption given it's static final), use "INTERNAL_TAGS" as the name.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+
+  // Takes and puts a list of BlogMessage type in map.
+  // Each BlogMessage type should contain tag that matches tags in the internal System. 
+  public static void putInMap(List<BlogMessage> messages, Map<String, LinkedList<BlogMessage>> map) {
```
peizhaoou, line 31: I think a BlogHashMap by design should internally maintain some data structure, which should be a "Map<String, LinkedList<BlogMessage>>", which then should be a private field of the BlogHashMap class. And then there are public non-static methods that client code can access/mutate the internal data structure.

We should not use public static method here and elsewhere in this class. Otherwise, there are really not abstraction here. As a concrete example, here there should probably be a method like:

public void putInMap(List<BlogMessage> messages {...}

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+    for (BlogMessage message : messages) {
```
peizhaoou, line 32: The nested for-loop here can use a private helper method like put() which just put one message in the internal data structure. That would make your code more readable.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+            // No such mapping with specified tag,
+            // Create first mapping with specified tag.
+            LinkedList<BlogMessage> firstValue = new LinkedList<BlogMessage>();
+            firstValue.addFirst(message); 
+            map.put(message.getTag(), firstValue);
+          }
+        } 
+      }
+      if (!isTagSupported) {
+        System.out.println("Tag not supported. Tag: " + message.getTag());
```
peizhaoou, line 49: Prefer "System.err.println" if this is intended to be an error message vs. an info message.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+        System.out.println("For BlogMessage@ " + message);
+      }
+    }
+
+  }
+
+  public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
```
peizhaoou, line 56: Same static method comment applies here.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+
+  // Takes map, and
+  // Returns a linkedList of all values of BlogMessage type in the map. 
+  public static LinkedList<BlogMessage> getMessages(Map<String, LinkedList<BlogMessage>> map) {
+    System.out.println("Fetching all [LinkedList] Values in map...");
+
+    Set<String> keys = map.keySet();
+    LinkedList<BlogMessage> values = new LinkedList<BlogMessage>();
+    for (String key : keys) {
+      values.addAll(map.get(key));
```
peizhaoou, line 68: What is supposed to be included in the returned linked list? There seems to be duplicate messages, would that be a problem? How would this be used?

If those are all well thought out, please put that in the method comment as well because it doesn't seem straightforward and worth extra expalantion.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    LinkedList<BlogMessage> values = getMessages(map);
+    if (limit > values.size()) {
+      limit = values.size();
+    } 
+
+    System.out.println("Fetching requested amount of [LinkedList] Values for all tags...");
+
+    Iterator<BlogMessage> iterateList = values.iterator();
```
peizhaoou, line 87: Line 87-94 looks a bit inefficient because you call getMessages in line 80, copying all the items to values already, and then you are trying to copy the portion within limit twice, and would have a waste in copying the rest in the first call to getMessages(). Try to see if you can improve that.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+  // Takes a tag, and
+  // Returns a linkedList of all values of BlogMessage type for the specified tag.
+  public static LinkedList<BlogMessage> getMessages(String tag, Map<String, LinkedList<BlogMessage>> map) {
+    System.out.println("Fetching all [LinkedList] Value(s) for specified tag...");
+
+    LinkedList<BlogMessage> values = new LinkedList<BlogMessage>();
+    if (map.containsKey(tag)) {
+      values.addAll(map.get(tag));
+    } else {
+      System.out.println("Tag not in map. Tag : " + tag);
```
peizhaoou, line 106: We probably don't want to add too much of info logs like these in the codebase. If they are only useful for one-time debugging, please remove them when trying to submit code.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  // Returns a requested amount of all BlogMessages for a specified tag.
+  public static LinkedList<BlogMessage>getMessages(String tag, Map<String, LinkedList<BlogMessage>> map, int limit) {
+    if (limit < 0) {
+      System.out.println("Cannot load this amount.");
+      return new LinkedList<>();
+    }
+
+    LinkedList<BlogMessage> values = getMessages(tag, map);
```
peizhaoou, line 118: Same here. We will be doing extra work that will be abandoned in the end. Please try to improve it.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+    if (limit > values.size()) {
+      limit = values.size();
+    } 
+
+    System.out.println("Fetching requested amount of [LinkedList] Values for specified tag...");
```
peizhaoou, line 123: Same for verbose print?

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  // Returns a requested amount of all BlogMessages for specified tags.
+  public static LinkedList<BlogMessage>getMessages(List<String> tags, Map<String, LinkedList<BlogMessage>> map, int limit) {
+    if (limit < 0) {
+      System.out.println("Cannot load this amount.");
+      return new LinkedList<>();
+    }
+
+    LinkedList<BlogMessage> values = getMessages(tags, map);
```
peizhaoou, line 158: Same here.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -75,18 +80,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         messages.add(message);
       }
 
+      // Create linkedHashMap and put in BlogMessages. 
+      Map<String, LinkedList<BlogMessage>> map = new LinkedHashMap<String, LinkedList<BlogMessage>>();
```
peizhaoou, line 84: This is why we want to make BlogHashMap a better abstraction. This right now basically exposes all internal details to the client code, which is particularly undesirable. A better abstraction would be that I can use the BlogHashMap like:

BlogHashMap blog_map = new BlogHashMap();
blog_map.putInMap(messages);

And as a client of your BLogHashMap, I don't and should not know anything about how you organize the internal data structure. A good standard is that I should not change a single line of my code if you decided that a hashmap is not great and you need to make an even fancier data structure.

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -23,16 +23,41 @@
   private final String sender;
   private final long timestamp;
   private final String image;
-  private final String tags;
+  private final String tag;
   private static ArrayList<String> messageReplies;
 
-  public BlogMessage(long id, String tags, String message, String image, String sender, ArrayList<String> messageReplies, long timestamp) {
+  public BlogMessage(long id, String tag, String message, String image, String sender, ArrayList<String> messageReplies, long timestamp) {
```
peizhaoou, line 29: Line too long? Please break it up.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Since each Blogmessage has one tag I don't think there'll be duplicate messages for a single tag. There might be same messages for different tags but that will not be a problem. I don't know if that clarifies what you addressed in the comment

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: fron what I understand you mean getting all items in the LinkedList and then copying a portion into the limitedLinkedList is double job, and the rest that are not copied will basically be useless?
So instead of first getting the original values and then copying part of it I could just get the requested amount straight from the hashmap. Is that close to addressing your comment?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: Understood

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
peizhaoou: Exactly! This usually matters when the HashMap is big, cause the time complexity is bound to the size of the internal hashmap, not the amount the caller requested, i.e., even if I just need one item (limit = 1), I would need to copy the whole hashmap, and then return the first one:)

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
peizhaoou: That clarifies it. That was what I thought (one message can have multiple tags) too.

In this case, I think it would be good to add a comment in the method saying "The returned list may contain duplicated messages" (just as an FYI for callers so that they don't accidentally think this is a list of unique items).

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Done!

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Done!

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Done!

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Done!

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Done!

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: Done!

PR #17, opened 2020-07-06
Implement LinkedHashMap for BlogMessages

Andrewmob2703-zz:


Andrewmob2703-zz:
Ignore first implementation of LinkedHashMap commited on July 1

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,44 +24,47 @@
 import java.util.Set;
 
 public final class BlogHashMap {
-  private static final String[] internalTags = {"#general", "#wellbeing", "#music", "#games"};
+  private final String[] INTERNAL_TAGS = {"#general", "#wellbeing", "#music", "#games"};
+  private Map<String, LinkedList<BlogMessage>> map;
```
peizhaoou, line 28: Nit: "map" is okay in this scenario since it should be the only hashmap in BlogHashMap for now. Going forward, it would be good to avoid variable names such as list, map, array, etc. because they are too general, and actual reader will know their they are "map" from their type already.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,44 +24,47 @@
...
+
+  // Returns whether a message's tag is supported internally.
+  private boolean tagIsSupported(BlogMessage message) {
+    boolean isTagSupported = false;
+    for (int i = 0; i < INTERNAL_TAGS.length; i++){
+      if (message.getTag().equals(INTERNAL_TAGS[i])){
+        isTagSupported = true;
```
peizhaoou: Consider replacing "isTagSupported = true;" to "return true;" to return early and convey a more clear message to reader: "as long as one tag is matched, it is supported." Also, that way, you don't need to define the "isTagSupported" local variable because you can simply return false when reaching the end of the function.

peizhaoou: Also, when you start to support more tags (which means the array becomes larger), you could consider to use a HashMap for the INTERNAL_TAGS array to check existence/matching in O(1) vs. O(n). You don't have to do it in this PR, but keep that in mind or add a TODO comment for next change, etc.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,44 +24,47 @@
...
+        if (map.containsKey(message.getTag())){
+          map.get(message.getTag()).addLast(message);
+        } else {
+          // No such mapping with specified tag,
+          // Create first mapping with specified tag.
+          LinkedList<BlogMessage> firstValue = new LinkedList<BlogMessage>();
+          firstValue.addFirst(message); 
+          map.put(message.getTag(), firstValue);
+        }
+      } else {
```
peizhaoou, line 46: Consider checking error case first and "continue" early to avoid another level of nesting for line 45-54.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,44 +24,47 @@
...
-    return map.size();
   }
-
-  // Takes map, and
-  // Returns a linkedList of all values of BlogMessage type in the map. 
-  public static LinkedList<BlogMessage> getMessages(Map<String, LinkedList<BlogMessage>> map) {
-    System.out.println("Fetching all [LinkedList] Values in map...");
-
+  
+  /** FOR ALL getMessages() METHODS:
```
peizhaoou: Consider using one style of comment for the same method, pick one of /**/ or //.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,44 +24,47 @@
...
-
+  
+  /** FOR ALL getMessages() METHODS: 
+        In a case where different users assign different tags to equivalent messages,
+        Returned values could have duplicate messages - but for different tags.
+    */
+
+  // Returns a linkedList of all values of BlogMessage type in the map.
+  public LinkedList<BlogMessage> getMessages() {
     Set<String> keys = map.keySet();
```
peizhaoou: Consider calling the HashMap's values() method to iterate the values directly instead of using keySet() and get()

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -71,105 +74,126 @@ public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
...
+    LinkedList values = new LinkedList();
 
-    Iterator<BlogMessage> iterateList = values.iterator();
-    LinkedList<BlogMessage> limitedValues = new LinkedList<BlogMessage>();
-    int i = 0;
-    while (iterateList.hasNext() && i < limit){
-      limitedValues.addLast(iterateList.next());
-      i++;
+    for (String key : keys) {
+      if (limit <= 0) {break;}
```
peizhaoou: Nit: put single-line statement within {} into a new line.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -71,105 +74,126 @@ public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
...
+      Iterator iterator;
+      int tagMessagesSize = (map.get(key)).size();
```
peizhaoou: Nit: no need for the extra parenthesis => map.get(key).size()

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -71,105 +74,126 @@ public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
...
-    int i = 0;
-    while (iterateList.hasNext() && i < limit){
-      limitedValues.addLast(iterateList.next());
-      i++;
+    for (String key : keys) {
+      if (limit <= 0) {break;}
+      Iterator iterator;
+      int tagMessagesSize = (map.get(key)).size(); 
+
+      // startPoint is from (n0.of.messages in each tag - limit) if the list contains more messages
```
peizhaoou: Nit: the comment is kind of confusing. Particularly, what is n0.of.messages?

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -71,105 +74,126 @@ public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
...
+      // than the limit,
+      // That way, returned values contain the most recent messages.
+      if (tagMessagesSize > limit) {
+        int startPoint = tagMessagesSize - limit;
+        iterator = (map.get(key)).listIterator(startPoint);
+      } 
+      // Otherwise the n0.of messages in the tag is less than or equal to the limit, hence
+      // we return all of them.
+      else {
```
peizhaoou: Nit: please do not break a "} else {" with comments because it's easy to misread. If the comments in line 98-99 are to describe the else branch, consider:

} else {
  // Otherwise the ...
  // ...
}

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -71,105 +74,126 @@ public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
   }
 
   // Returns a requested amount of all BlogMessages in map
-  public static LinkedList<BlogMessage> getMessages(Map<String, LinkedList<BlogMessage>> map, int limit) {
+  // Returned values contain the most recent messages.
```
peizhaoou: Nit: it would be good to give more details about the argument "limit". I am confused here because I am not entirely sure if "limit" here is to limit the size of the returned LinkedList? E.g., if limit=10, does it mean that the returned list will have at most 10 elements? Please be explicit in the comment.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -71,105 +74,126 @@ public static int getSize(Map<String, LinkedList<BlogMessage>> map) {
...
+    for (String key : keys) {
+      if (limit <= 0) {break;}
+      Iterator iterator;
+      int tagMessagesSize = (map.get(key)).size(); 
+
+      // startPoint is from (n0.of.messages in each tag - limit) if the list contains more messages
+      // than the limit,
+      // That way, returned values contain the most recent messages.
+      if (tagMessagesSize > limit) {
+        int startPoint = tagMessagesSize - limit;
```
peizhaoou: "startPoint" is a local variable that is used only once right next line. Would be good to just pass "tagMassageSize - limit" to listIterator() next line.

peizhaoou: Same applies else where.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
martiKallas: Just to add to this, it seems that // comments are preferred over multi-line comments in Google code.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,199 @@
...
+        values.addAll(map.get(tag));
+      } else {
+        System.err.println("Tag not found. Tag : " + tag);
+      }    
+    }
+    return values;
+  }
+
+  // Returns a requested amount of all BlogMessages for specified tags.
+  public LinkedList<BlogMessage> getMessages(List<String> tags, int limit) {
```
martiKallas, line 114: You only really need this function. If tags is empty or null, get all tags. If limit <= 0 don't use a limit. Just document that in the comments of the function. This way you reuse a lot of code and you limit the number of functions that need to be tested.

This also reduces the PR size to a more reasonable level.

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -23,16 +23,40 @@
   private final String sender;
   private final long timestamp;
   private final String image;
-  private final String tags;
+  private final String tag;
   private static ArrayList<String> messageReplies;
 
-  public BlogMessage(long id, String tags, String message, String image, String sender, ArrayList<String> messageReplies, long timestamp) {
+  public BlogMessage(long id, String tag, String message, String image, String sender, ArrayList<String> messageReplies, long timestamp) {
```
martiKallas: I would keep this in its own file since it is used in other parts of the code as well, and the uses of BlogMessage aren't dependent on your hashtable API. For example, storing a BlogMessage in Datastore, doesn't also require that Datastore knows about the getMessage function.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -74,19 +79,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         BlogMessage message = new BlogMessage(messageId, tags, comment, image, sender, messageReplies, timestamp);
         messages.add(message);
       }
+
```
martiKallas: Normally, linking this into the other parts of the code would be done in a separate PR. For this case it would be for two reasons: to make the PR smaller and more focused on a single goal (implementation), and because this code isn't tested yet. This would be a perfect chance to practice branching a new change off of your current change and migrating this code to that new branch.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public final class BlogHashMap {
+  private final String[] INTERNAL_TAGS = {"#general", "#wellbeing", "#music", "#games"};
```
martiKallas: You might consider making these tags a public constant in another file if they are going to be used in a number of places.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Got it

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: sounds good

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Got it

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Sounds good!

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: couldn't find it

Andrewmob2703-zz: Oh just found it in the doc. It's the last method in the doc so my eyes missed it lol

.gitignore
```diff
@@ -0,0 +1,2 @@
+target/
```
martiKallas: In cases like this, it is better to create tiny PR's. .gitignore doesn't have anything to do with this PR and a tiny change can be reviewed very quickly.

martiKallas (changes requested):
Make sure you get a review from either Steven or Tayyaba!

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -46,14 +47,21 @@
...
 /** Servlet that returns some example content. TODO: modify this file to handle comments data */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
     int numberOfCommentsToDisplay = 0;
     
     @Override
-    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {      
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException { 
+      //Get number of comments.
+      numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
```
smarquez-jpg, line 59: May not be a problem but I know that when I got the number of comments to display in the doGet instead of the doPost, I ran into the problem where the variable would not update. Other than that LGTM

smarquez-jpg (approved):

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -147,39 +104,32 @@ public void putInMap(List<BlogMessage> messages) {
...
+      return getMessages(tags.get(0), limit);
     }
 
     LinkedList values = new LinkedList();
-
     for (String tag : tags) {
-      if (limit <= 0) {break;}
-      Iterator iterator;
+      if (limit <= 0) 
+      {break;}
```
peizhaoou: Nit: use the recommended style guide for consistency:

if (limit <= 0) {
  break;
}

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -147,39 +104,32 @@ public void putInMap(List<BlogMessage> messages) {
...
-
-  // Returns a requested amount of all BlogMessages for specified tags.
+  // This method takes a list of tags and a number of messages to load, and
+  // Returns a requested (or default) amount of BlogMessages under specified tag/tags.
+  // If tag/tags are not specified, a requested (or default) amount of all BlogMessages will be returned.
   public LinkedList<BlogMessage> getMessages(List<String> tags, int limit) {
-    if (limit < 0) {
-      return new LinkedList<>();
+    if (limit <= 0) {
+      limit = DEFAULT_LOAD_AMOUNT;
```
peizhaoou, line 116: Awesome, good to see you use a constant this way in this case!

portfolio/src/main/java/com/google/sps/data/InternalTags.java
```diff
@@ -0,0 +1,53 @@
...
+
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class InternalTags {
+  private final List<String> INTERNAL_TAGS = new ArrayList<String>();
+  private final String DEFAULT_TAG = "#general";
+  private final void setTags() {
```
peizhaoou: Any reason we need to make the methods like setTags() & tagsIsSupported() final?

portfolio/src/main/java/com/google/sps/data/InternalTags.java
```diff
@@ -0,0 +1,53 @@
...
+    for (String internalTag : INTERNAL_TAGS) {
+      if (message.getTag().equals(internalTag)){
+        return true;
+      }
+    }
+    return false;
+  }
+
+  // For check in DataServlet
+  public String defaultTag(){
```
peizhaoou: This could be a static method because they are not specific to any InternalTags object, but a shared method so that one can call "InternalTags.defaultTag()" without instantiating an InternalTags object.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -25,14 +25,13 @@
...
   private final int DEFAULT_LOAD_AMOUNT = 10;
-  private final InternalTags internal_tags;
   private Map<String, LinkedList<BlogMessage>> map;
 
   // Takes and puts a list of BlogMessage type in map.
   // Each BlogMessage type should contain tag that matches tags in the internal System. 
   public void putInMap(List<BlogMessage> messages) {
     for (BlogMessage message : messages) {
-      if (internal_tags.tagIsSupported(message)) {
+      if (InternalTags.tagIsSupported(message)) {
```
peizhaoou, line 38: Would checking the opposite make the nesting better, e.g.:

if (!InternalTags.tagIsSupported(message)) {
  continue;
}
// Process the case with supported tag with 1 fewer level of nesting.

peizhaoou:
LGTM overall. Please make sure that the current BlogHashMap pass the tests Steven added before you merge.

.gitignore
peizhaoou: Echo Marti's point. I am fine with leaving it as is for this PR. But going forward, if reviewers make a suggestion like this and there's no strong reason for not doing so, the right way to really address the comment is to revert the change, break it up, and send it for review in a separate PR.

peizhaoou: Keep in mind that having smaller and more focused PRs is really a good practice of separation of concerns. It would make your and reviewer's life much easier with focused changes because you want your reviewers to be very focused, and you don't want to deal with too many comments all at once.

peizhaoou (approved):

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
martiKallas: I like how you organized this with the private getMessages functions serving internal purposes!

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,150 @@
...
+      System.err.println("Tag not found. Tag: " + tag);
+    }
+    return values;
+  } 
+
+  // This method takes a list of tags and a number of messages to load, and
+  // Returns a requested (or default) amount of BlogMessages under specified tag/tags.
+  // If tag/tags are not specified, a requested (or default) amount of all BlogMessages will be returned.
+  public LinkedList<BlogMessage> getMessages(List<String> tags, int limit) {
+    if (limit <= 0) {
```
martiKallas, line 115: I would recommend against this. AFAIK, it isn't a common convention that if limit <= 0, then choose some other number in it's place. It is slightly more common that if limit == 0 then use no limit. 

Further, DEFAULT_LIMIT doesn't mean much from the user's point of view. I think in this case it is better to imagine you are writing a library that will be used by someone other than your pod. Then ask if they would want a default value and why.

Always try to make sure default values are meaningful rather than arbitrary. If you can't come up with a meaningful default value, then it is usually better to handle that situation in another way like throwing an error, or re-defining the function.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -0,0 +1,150 @@
...
+      }
+    } else {
+      System.err.println("Tag not found. Tag: " + tag);
+    }
+    return values;
+  } 
+
+  // This method takes a list of tags and a number of messages to load, and
+  // Returns a requested (or default) amount of BlogMessages under specified tag/tags.
+  // If tag/tags are not specified, a requested (or default) amount of all BlogMessages will be returned.
```
martiKallas: nit: At least for C++, the common style is to bracket a parameter name with vertical slashes: If |tags| are not specified...

martiKallas (approved):
I think this is really nice work! I just left a couple of minor comments.

I think you can do a little better with splitting things up into smaller PRs as well. Each file you changed could have been it's own PR, although combining InternalTags and BlogHashMap makes sense.

PR #18, opened 2020-07-06
Add nickname

tayyaba17:
I added the nickname capability but when the user clicks the submit button, it redirects them to "/home" and I can't figure out where to change that... Let me know if you see anything that may help

### DIFF
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -28,6 +28,8 @@ import com.google.appengine.api.datastore.Query.SortDirection;
 import com.google.appengine.api.images.ImagesService;
 import com.google.appengine.api.images.ImagesServiceFactory;
 import com.google.appengine.api.images.ServingUrlOptions;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
 import com.google.gson.Gson;
 import com.google.sps.data.BlogMessage;
 import java.io.IOException;
@@ -63,15 +65,18 @@ public class DataServlet extends HttpServlet {
       DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
       PreparedQuery results = datastore.prepare(query);
 
+      UserService userService = UserServiceFactory.getUserService();
+
       for (Entity entity : results.asIterable()) {
         long messageId = entity.getKey().getId();
         long timestamp = (long) entity.getProperty("time");
         String tags = (String) entity.getProperty("tag");
         String comment = (String) entity.getProperty("text");
-        String sender = (String) entity.getProperty("sender");
+        String nickname = (String) entity.getProperty("nickname");
+        String email = (String) userService.getCurrentUser().getEmail();
         String image = (String) entity.getProperty("imgUrl");
         ArrayList<String> messageReplies = (ArrayList) entity.getProperty("replies");
-        BlogMessage message = new BlogMessage(messageId, tags, comment, image, sender, messageReplies, timestamp);
+        BlogMessage message = new BlogMessage(messageId, tags, comment, image, nickname, email, messageReplies, timestamp);
         messages.add(message);
       }
 
@@ -108,7 +113,7 @@ public class DataServlet extends HttpServlet {
       //Get number of comments.
       numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
 
-      // Get the URL of the image that the user uploaded to Blobstore.
+      // Get the URL of  the image that the user uploaded to Blobstore.
       String imageUrl = getUploadedFileUrl(request, "image");
 
       //Get replies.
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -101,31 +101,23 @@ public class DataServlet extends HttpServlet {
    */
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-      // Get the message entered by the user.
       String message = request.getParameter("text-input");
 
-      // Get sender.
       String sender = getParameter(request, "sender", "Steven");
 
-      // Get type of comment.
       String commentType = getParameter(request, "tags", "Default");
       
-      //Get number of comments.
       numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
 
-      // Get the URL of  the image that the user uploaded to Blobstore.
       String imageUrl = getUploadedFileUrl(request, "image");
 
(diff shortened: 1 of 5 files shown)


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -108,7 +113,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       //Get number of comments.
       numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
 
-      // Get the URL of the image that the user uploaded to Blobstore.
+      // Get the URL of  the image that the user uploaded to Blobstore.
```
peizhaoou: Nit: redundant space?

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -18,24 +25,68 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect("/login");
+      return;
+    }
+
+    String nickname = request.getParameter("nickname");
```
peizhaoou: For string constants like "nickname", it would be good to make it a constant for better reuse, e.g.:

private final String NICKNAME = "nickname";

and then use NICKNAME here.

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -18,24 +25,68 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    // The put() function automatically inserts new data or updates existing data based on ID
+    datastore.put(entity);
 
+    response.sendRedirect("/");
+  }
+
+  /**
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
+   */
+  private String getUserNickname(String id) {
```
peizhaoou, line 84: Is the function argument "id" never used?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -108,7 +113,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       //Get number of comments.
```
martiKallas: nit: you probably don't need so many comments. Comments are good, but they are normally used for describing APIs/functions, describing code that isn't clear, or explaining choices that aren't clear. 

This comment (and some others in this file) isn't really needed because the code is easy to understand by reading the variable and function names. This is called 'self-documenting' code.

martiKallas (commented):
For the problem with redirecting to the home page, try thinking through the execution path. I'm not sure if this is the actual problem, but I think it might be. Where does the user get directed when they hit the submit nickname button? Where does that code direct them?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
tayyaba17: I figured the redirect problem out earlier! It was redirecting it to an old file that I decided to remove. Will remove extra comments, thanks!

tayyaba17:
I just updated some of the files based on the comments. I also fixed the redirect error. I think the authentication portion is set for now. When we work on displaying blogposts we should display the nickname of the user.

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -34,8 +34,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       String userEmail = userService.getCurrentUser().getEmail();
 
       String nickname = getUserNickname(userService.getCurrentUser().getUserId());
+    
+      if (nickname == ""){ out.println("<p>You don't have a nickname</p>");
```
peizhaoou: A couple comments:

(1) for nickname == "", are you supposed to check emptiness? If so, I think you should use nickname.isEmpty(). This right now is comparing reference equality.

(2) Use a new line after {, and add a space between ){.

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -34,8 +34,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      } else { out.println("<p>Your current nickname is " + nickname + ". If you'd like to change it, enter a new one below.</p>");}
```
peizhaoou: Same here: start a new line.

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -60,33 +64,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       return;
     }
 
-    String nickname = request.getParameter("nickname");
+    String NICKNAME = request.getParameter("nickname");
```
peizhaoou: Sorry I was probably confusing for using constants. I meant having NICKNAME as a class member and use that here, e.g.:

private final static String NICKNAME = "nickname";

In the function body, use it like request.getParameter(NICKNAME)

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -60,33 +64,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        new Query("blogMessage")
+            .setFilter(new Query.FilterPredicate("id", Query.FilterOperator.EQUAL, id));
     PreparedQuery results = datastore.prepare(query);
     Entity entity = results.asSingleEntity();
     if (entity == null) {
       return "";
     }
-    String nickname = (String) entity.getProperty("nickname");
-    return nickname;
+    String NICKNAME = (String) entity.getProperty("nickname");
```
peizhaoou: Same here.

portfolio/src/main/java/com/google/sps/servlets/LoginServlet.java
tayyaba17: so after I have the class member NICKNAME, would the variables inside the functions be named "nickname" ? What's the best way to do that?

peizhaoou: So per the Java style guide, constant variables would be named with everything capitalized. It's just a variable, but it's like a class-level global variable that you can refer to in your code. The motivation behind this is that if someday you need to change the parameter name from "nickname" to "alias", you can just change one place.

peizhaoou (approved):
Thanks Tayyaba!

PR #20, opened 2020-07-07
Steven test with hash map

smarquez-jpg:


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -77,7 +77,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         String sender = (String) entity.getProperty("sender");
         String image = (String) entity.getProperty("imgUrl");
         ArrayList<String> messageReplies = (ArrayList) entity.getProperty("replies");
-        BlogMessage message = new BlogMessage(messageId, tags, comment, image, sender, messageReplies, timestamp);
+        BlogMessage message = new BlogMessage(messageId, tags, comment, image, sender, "", messageReplies, timestamp);
```
Andrewmob2703-zz: you can probably resolve conflicts so that you get the user email and use that for merging instead of the empty string.

Andrewmob2703-zz:
I think you can resolve conflicts so that you get the userEmail for where you currently have an empty string in the DataServlet

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
peizhaoou: A general tip: if we need to pass a constant (String, int, double, bool, etc.) to a function argument, it's usually good to use an inlined comment like

BlogMessage message = new BlogMessage(
    messageId, tags, comment, image, sender, /*email=*/"", messageReplies, timestamp);

Then reader knows what argument is getting the "" empty string.

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -43,9 +43,9 @@ public void oneBlogPostForEachTag() {
     //Return a Hashmap with LinkedList of size one for each tag.
     BlogHashMap actual = new BlogHashMap();
     List<BlogMessage> temp = new ArrayList<>();
-    temp.add(new BlogMessage((long) 1, "#general", "I enjoy pop music", "", "Steven", new ArrayList<String>(), System.currentTimeMillis()));
-    temp.add(new BlogMessage((long) 2, "#wellbeing", "Country music is the best music", "", "Tayyaba", new ArrayList<String>(), System.currentTimeMillis()));
-    temp.add(new BlogMessage((long) 3, "#music", "Classical music makes you smarter", "", "Andrew", new ArrayList<String>(), System.currentTimeMillis()));
+    temp.add(new BlogMessage((long) 1, "#general", "I enjoy pop music", "", "Steven", "", new ArrayList<String>(), System.currentTimeMillis()));
```
peizhaoou: Inlined comments applies here for all the constants and elsewhere as well.

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -43,9 +43,9 @@ public void oneBlogPostForEachTag() {
...
+    temp.add(new BlogMessage((long) 2, "#wellbeing", "Country music is the best music", "", "Tayyaba", "", new ArrayList<String>(), System.currentTimeMillis()));
+    temp.add(new BlogMessage((long) 3, "#music", "Classical music makes you smarter", "", "Andrew", "", new ArrayList<String>(), System.currentTimeMillis()));
```
peizhaoou: We definitely need to break the long line (elsewhere as well). The line limit is 100 characters for Java.

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -100,7 +100,7 @@ public void invalidTag(){
     //Tag should not be supported so should return empty hash table.
     BlogHashMap actual = new BlogHashMap();
     List<BlogMessage> temp = new ArrayList<>();
-    temp.add(new BlogMessage(1, "#Edm", "I listen to Edm while I work.", "", "Steven", new ArrayList<String>(), System.currentTimeMillis()));
+    temp.add(new BlogMessage((long) 1, "#Edm", "I listen to Edm while I work.", "", "Steven", "", new ArrayList<String>(), System.currentTimeMillis()));
```
peizhaoou: For long constants, you don't need the (long) conversion, just append a "l" character after the number to tell the compiler to use a long integer here.

portfolio/src/test/java/com/google/sps/HashTableTest.java
peizhaoou: "temp" in general is really not a good name for variables. Consider using "messages" or "blogMessages"

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -61,13 +61,13 @@ public void multipleBlogPostsForFewTags() {
     BlogHashMap actual = new BlogHashMap();
     List<BlogMessage> temp = new ArrayList<>();
     for(int i = 0; i < 4; i++){
```
peizhaoou: Nit: Put a space between ) & {

peizhaoou: Nit: This is me nitpicking and I do not have strong feeling (especially in testing code), so your call.

I see that the constant 4 is used four times. It would be good to use a local variable like:

int iterations = 4;

And then use the variable.

portfolio/src/test/java/com/google/sps/HashTableTest.java
```diff
@@ -39,60 +41,66 @@ public void noComments() {
...
+    temp.add(new BlogMessage((long) 1, "#general", "I enjoy pop music", "", "Steven", new ArrayList<String>(), System.currentTimeMillis()));
+    temp.add(new BlogMessage((long) 2, "#wellbeing", "Country music is the best music", "", "Tayyaba", new ArrayList<String>(), System.currentTimeMillis()));
+    temp.add(new BlogMessage((long) 3, "#music", "Classical music makes you smarter", "", "Andrew", new ArrayList<String>(), System.currentTimeMillis()));
+    actual.putInMap(temp);
 
-    Set<String> keys = actual.keySet();
+    Collection<String> keys = Arrays.asList("#general", "#wellbeing", "#music");
     for(String key : keys){
-      Assert.assertEquals(1, actual.get(key).size());
+      Assert.assertEquals(1, actual.getMessages(key).size());
```
peizhaoou, line 56: About checking the equality of two objects x & y, if you define a equals() method in the object's class, you can just use Assert.assertEquals(x, y);

This would be more convenient, expressive, and concise.

peizhaoou (commented):
Thanks Steven! Great to see the testing integration is working. LGTM overall, just left some comments. I think submitting after addressing them is fine with me, but I just wonder if this would need to wait for the implementation to be checked in for this to be run?

peizhaoou (approved):
Thanks Steven! Great to see the testing integration is working. LGTM overall, just left some comments. I think submitting after addressing them is fine with me, but I just wonder if this would need to wait for the implementation to be checked in for this to be run?

.gitignore
```diff
@@ -0,0 +1,2 @@
+target/
```
martiKallas, line 1: Before you submit this, make a single, separate change (PR) to add the .gitignore file and then merge that into this PR. That way you won't be including unneeded files.

You might be able to do that directly from the Github interface.

martiKallas (dismissed):

martiKallas:
Please get a review from either Andrew or Tayyaba as well!

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -20,18 +20,20 @@
 public final class BlogMessage {
   private final long id;
   private final String message;
-  private final String sender;
+  private final String nickname;
   private final long timestamp;
   private final String image;
-  private final String tag;
+  private final String email;
+  private final String tags;
```
Andrewmob2703-zz: We used 'tag' because each post can only have one tag. Currently there's no support for multiple tags per post. Please change that before merging.

Andrewmob2703-zz (approved):
Looks good! Just change tags to tag since we're only supporting one tag per post

PR #28, opened 2020-07-10
Fixed bug where nickname and image would not display

smarquez-jpg:


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -54,16 +54,8 @@
...
-      numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
-           
-      if (numberOfCommentsToDisplay < 1 || numberOfCommentsToDisplay > 100) {
-        response.setContentType("text/html");
-        response.getWriter().println("Please enter an integer between 1 and 100.");
-        return;
-      }
-
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {      
+      System.out.println(numberOfCommentsToDisplay);
```
tayyaba17: Just a reminder to remove this when you're done testing.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -54,16 +54,8 @@
     int numberOfCommentsToDisplay = 0;
     
     @Override
-    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException { 
-      //Get number of comments.
-      numberOfCommentsToDisplay = getNumberOfCommentsToDisplay(request);
```
tayyaba17, line 59: Did you delete this for now and then you'll add another way to receive the user's # of comment preference next?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -129,7 +150,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       long timestamp = System.currentTimeMillis();
 
       Entity blogMessageEntity = new Entity("blogMessage");
-      blogMessageEntity.setProperty("sender", sender);
+      blogMessageEntity.setProperty("nickname", sender);
```
tayyaba17: Not an urgent thing, but I also think we should update instances of sender through out the other files so that it's not confusing. Sometimes in data store there are separate columns for sender and for nickname when there should just be one.

tayyaba17 (commented):
I think it looks good! Thanks for getting the nickname to show next to the post

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,6 +24,7 @@
 import java.util.Set;
 
 public final class BlogHashMap {
+  private final String[] INTERNAL_TAGS = {"#general", "#wellbeing", "#music", "#games", "Tags"};
```
martiKallas, line 27: Can you explain what "Tags" is for and why it doesn't follow the same pattern as the others? This is a perfect time to use a code comment. You might also consider making it a separate constant if it doesn't fit the pattern of INTERNAL_TAGS.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -24,7 +24,7 @@
 import java.util.Set;
 
 public final class BlogHashMap {
-  private final String[] INTERNAL_TAGS = {"#general", "#wellbeing", "#music", "#games"};
+  private final String[] INTERNAL_TAGS = {"#general", "#wellbeing", "#music", "#games", "Tags"};
```
peizhaoou, line 27: Why is there tag name "Tags" which looks inconsistent with other tags?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
martiKallas: +1 - for the most part, avoid print statements in production code. You can use them for debugging in prod, but this is usually reserved for key steps or unexpected errors. In that case, you should include a detailed message to explain the value you are printing.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
smarquez-jpg: That was just there as a default tag since my branch didn't have the dropdown menu

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -89,7 +81,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       BlogHashMap blogMap = new BlogHashMap();
       blogMap.putInMap(messages);
 
-      // Load messages from BlogHashMap and respond with gson.
+      // If (user loads all BlogMessages) 
+      LinkedList<BlogMessage> allBlogMessages = blogMap.getMessages();
+
+      if (numberOfCommentsToDisplay < 1 || numberOfCommentsToDisplay >= allBlogMessages.size()) {
```
martiKallas: From a usability standpoint, I think it would be better to set numberOfCommentsToDisplay = allBlogMessages.size() if the user is requesting more comments than are available. The user is likely to want the max number of comments anywas, and this saves the user a round trip of input and response.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -110,6 +111,13 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       if(numberOfCommentsToDisplay == 0){
         response.getWriter().println(gson.toJson(allBlogMessages)); // set a default amount later.
         return;
+      } else {
+        List<BlogMessage> limitedBlogMessages = new ArrayList<>();
+        for (int i = 0; i < allBlogMessages.size(); i++) {
+          limitedBlogMessages.add(allBlogMessages.get(i));
+        }
+        response.getWriter().println(gson.toJson(limitedBlogMessages));
+        return;
```
peizhaoou: Both the if and else branches will return, which means the code after line 121 would not be reachable. Is that intended? Normally, unreachable code should be deleted, but I assume you add things here for a quick fix, so I just ask in case.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -89,7 +81,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       BlogHashMap blogMap = new BlogHashMap();
       blogMap.putInMap(messages);
 
-      // Load messages from BlogHashMap and respond with gson.
+      // If (user loads all BlogMessages) 
+      LinkedList<BlogMessage> allBlogMessages = blogMap.getMessages();
```
martiKallas: This uses an old version of the BlogHashMap API, please use the current public method. If you think something is missing from the API comments, please ask Andrew to clarify them.

There are also a few below I noted.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -89,7 +81,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+      if (numberOfCommentsToDisplay < 1 || numberOfCommentsToDisplay >= allBlogMessages.size()) {
+        response.setContentType("text/html");
+        response.getWriter().println("Please enter an integer between 1 and "+allBlogMessages.size()+".");
+        return;
+      }
+
+      // If (user loads BlogMessages for a specific tag)
+      String tagToSearch = ""; // we'll get the input later.
+      LinkedList<BlogMessage> BlogMessagesForTag = blogMap.getMessages(tagToSearch);
```
martiKallas: This is an old API function.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -89,7 +81,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      LinkedList<BlogMessage> allBlogMessages = blogMap.getMessages();
+
+      if (numberOfCommentsToDisplay < 1 || numberOfCommentsToDisplay >= allBlogMessages.size()) {
+        response.setContentType("text/html");
+        response.getWriter().println("Please enter an integer between 1 and "+allBlogMessages.size()+".");
+        return;
+      }
+
+      // If (user loads BlogMessages for a specific tag)
+      String tagToSearch = ""; // we'll get the input later.
```
martiKallas: This will cause a bug. I want you to try and track down what the error would be because I think it is a good exercise in using/reading APIs. Check the API documentation and if something isn't clear, ask Andrew to clarify the comments.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
peizhaoou: It seems for that purpose, it would still be good to have something like "#default_tags" for consistency. But your call.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -98,8 +103,24 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       response.setContentType("application/json;");
       
-      response.getWriter().println(gson.toJson(loadedBlogMessages));
-      return;
+      if(numberOfCommentsToDisplay == 0){
+        response.getWriter().println(gson.toJson(allBlogMessages)); // set a default amount later.
+        return;
+      } else {
+        List<BlogMessage> limitedBlogMessages = new ArrayList<>();
+        for (int i = 0; i < allBlogMessages.size(); i++) {
```
martiKallas: This seems to copy allBlogMessages into limitedBlogMessages. I think you meant to use numberOfCommentsToDisplay?

peizhaoou (approved):
LGTM and please resolve remaining issues. BTW, is this bug related to the one you and Tayyaba debugged yesterday?

martiKallas (dismissed):
It looks like you are using the old BlogHashMap API. Please fix that, and make sure you manually test your changes since we don't have unit tests.

PR #30, opened 2020-07-14
Fix logical error in BlogHashMap

Andrewmob2703-zz:
**Specifically:** Fixed bug where if we wanted 1 message the returned message was the last message for the first tag in the map, and not what we want - which is the last message for the last tag entered.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
+++ b/portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
@@ -20,6 +20,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
 
@@ -36,7 +37,7 @@ public final class BlogHashMap {
   public void putInMap(List<BlogMessage> messages) {
     for (BlogMessage message : messages) {
       if (InternalTags.tagIsSupported(message)) {
-        if (!map.containsKey(message.getTag())){
+        if (!map.containsKey(message.getTag())) {
           LinkedList<BlogMessage> firstValue = new LinkedList<BlogMessage>();
           firstValue.addFirst(message); 
           map.put(message.getTag(), firstValue);
@@ -60,26 +61,26 @@ public final class BlogHashMap {
 
   // Returns a requested amount of all BlogMessages in map
   private LinkedList<BlogMessage> getMessages(int limit) {
-    LinkedList values = new LinkedList();
+    LinkedList<BlogMessage> values = new LinkedList<BlogMessage>();
 
-    for (LinkedList<BlogMessage> tagMessages : map.values()){
+    // |valuesList| is a list of LinkedList BlogMessages for each tag in map, hence
+    // We reverse iterate starting from the last tag entered.
+    List valuesList = new ArrayList(map.values());
+    ListIterator reverseItr = valuesList.listIterator(valuesList.size());
+     
+    while (reverseItr.hasPrevious()) {
       if (limit <= 0) {
         break;
       }
 
-      Iterator iterator;
-      int tagMessagesSize = tagMessages.size();
+      LinkedList tagMessages = (LinkedList) reverseItr.previous();
 
-      // If a tag contains more messages than requested
-      // Iterate from (number of messages - limit) to return the most recent BlogMessages. 
-      if (tagMessagesSize > limit) {
-        iterator = tagMessages.listIterator(tagMessagesSize - limit);
-      } else {
-        iterator = tagMessages.iterator();
-      }
-
-      while(iterator.hasNext() && limit >= 0) {
-        values.addLast(iterator.next());
+      // Use LinkedList's |descendingIterator()| to iterate BlogMessages from the back,
+      // And addFirst to return the most recent messages in the order they were received.
+      // This way, if [limit] is 1 we return the last message in the last tag entered. 
+      Iterator iterateFromBack = tagMessages.descendingIterator();
+      while(iterateFromBack.hasNext() && limit > 0) {
+        values.addFirst((BlogMessage) iterateFromBack.next());
         limit--;
       }
     }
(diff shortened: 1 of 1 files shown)


portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -88,18 +89,21 @@ public void putInMap(List<BlogMessage> messages) {
 
   // Returns a requested amount of all BlogMessages for a specified tag.
   private LinkedList<BlogMessage> getMessages(String tag, int limit) {
```
tayyaba17, line 89: Is it best to have functions with the same name? I know since the parameters are different it shouldn't cause any errors, just not sure about convention.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -122,7 +126,7 @@ public void putInMap(List<BlogMessage> messages) {
       return getMessages(tags.get(0), limit);
     }
 
-    LinkedList values = new LinkedList();
+    LinkedList<BlogMessage> values = new LinkedList<BlogMessage>();
     for (String tag : tags) {
       if (limit <= 0) {
```
tayyaba17, line 129: I feel like this line is repeated a lot, not sure if it's worthwhile to just make a function that returns a bool based on the conditional, but worth considering

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -60,26 +61,26 @@ public void putInMap(List<BlogMessage> messages) {
...
   // Returns a requested amount of all BlogMessages in map
   private LinkedList<BlogMessage> getMessages(int limit) {
-    LinkedList values = new LinkedList();
-
-    for (LinkedList<BlogMessage> tagMessages : map.values()){
+    LinkedList<BlogMessage> values = new LinkedList<BlogMessage>();
+
+    // |valuesList| is a list of LinkedList BlogMessages for each tag in map, hence
+    // We reverse iterate starting from the last tag entered.
+    List valuesList = new ArrayList(map.values());
```
peizhaoou: Use the complete type List<List<String>> to make the type clear so that you don't need the comments in lines 66-67?

peizhaoou: My bad, I meant List<LinkedList<String>>

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -60,26 +61,26 @@ public void putInMap(List<BlogMessage> messages) {
...
+    ListIterator reverseItr = valuesList.listIterator(valuesList.size());
```
peizhaoou: Similarly use ListIterator<LinkedLists<String>> reverseItr so that you don't need the type conversion in line 76.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -137,7 +141,7 @@ public void putInMap(List<BlogMessage> messages) {
           iterator = map.get(tag).iterator();
         }
         while(iterator.hasNext() && limit >= 0) {
-          values.addLast(iterator.next());
+          values.addLast((BlogMessage) iterator.next());
```
peizhaoou: Similarly, give iterator a template type (i..e, a BlogMessage) when you declare the "iterator" to avoid the conversion.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
peizhaoou: Okay I made two mistakes in a row. For map.values(), I think it should be List<LinkedList<BlogMessage>>.

peizhaoou (commented):
Can you help me understand why the order matters?

Plus, I believe Java HashMap provides no guarantees about the order of its keys/values, i.e., map,values() won't return a guaranteed order[1].

[1] https://stackoverflow.com/questions/2144776/is-the-order-of-values-retrieved-from-a-hashmap-the-insertion-order/2144800#:~:text=From%20the%20Javadoc%3A%20HashMap%20%22class,TreeMap%20(for%20comparision%20order). This means your change here is not guaranteed to work?

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Yeah I think it's not a problem. We're just overloading them

Andrewmob2703-zz:
> Can you help me understand why the order matters?
> 
> Plus, I believe Java HashMap provides no guarantees about the order of its keys/values, i.e., map,values() won't return a guaranteed order[1].
> 
> This means your change here is not guaranteed to work?

Our map implements a LinkedHashMap so the order of our keys/values are maintained. Order matters because we want to get the most recent messages.
Suppose users input tag in the following order today: #general, #sports, and #games. That means that in our map the first key is #general and so on... If we wanted just 1 message the next day, the previous implementation would return the last message for #general, which is not the last message entered. 
The current implementation would return the most recent message which is the last message for #games, since we're iterating from the last key/tag in the map.
Hope that clarifies things.

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
Andrewmob2703-zz: Not sure I fully understand your comment. Could you please expand?

tayyaba17: this line "limit <= 0"

Andrewmob2703-zz: oh you mean like a boolean function that checks if limit is <= 0?

portfolio/src/main/java/com/google/sps/data/BlogHashMap.java
```diff
@@ -122,7 +126,7 @@ public void putInMap(List<BlogMessage> messages) {
       return getMessages(tags.get(0), limit);
```
peizhaoou, line 124: Can we have a check in the very beginning like

if (!map.containsKey(tag)) {
  return getMessage(tags.get(0), limit);
}

to keep the main logic more focused?

BTW, you may need a comment for "tags.get(0)" to tell readers why using the first tag.

peizhaoou (commented):
Thanks Andrew for the explanation. I think if the order actually matters and looks to be a part of the BlogHashMap's contract, we should have unit tests to reflect that. Can you add a test or update existing tests to capture this?

peizhaoou (approved):

PR #33, opened 2020-07-17
MVP that allows to post and get recent post 

Andrewmob2703-zz:


smarquez-jpg (approved):
LGTM

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -47,18 +48,18 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Iterator;
```
martiKallas: Please follow alphabetical order for the imports.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -47,18 +48,18 @@
...
 
-/** Servlet that returns some example content. TODO: modify this file to handle comments data */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
     int numberOfCommentsToDisplay = 0;
+    List<String> tagsToSearch = new ArrayList<String>();
+    List<String> messageReplies = new ArrayList<String>(); // TODO: Handle replies later.
```
martiKallas: nit: TODO comments should be on their own line for easier readability.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -76,97 +70,88 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       String message = request.getParameter("text-input");
-
-      String sender = getParameter(request, "sender", "Steven");
+      String sender = request.getParameter("sender");
+      String loadFactor = request.getParameter("load_factor");
+      String postTag = getTagParameter(request, "tags", InternalTags.defaultTag());
       
-      // TODO: 
-      //      Get default tag from the InternalTags class and use that below.
+      //TODO: Handle image file sent with FormData.
```
peizhaoou: Please keep the comment style consistent. Add one space after //

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -71,108 +72,86 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         String comment = (String) entity.getProperty("text");
         String nickname = (String) entity.getProperty("nickname");
         String email = (String) userService.getCurrentUser().getEmail();
-        String image = (String) entity.getProperty("imgUrl");
-        ArrayList<String> messageReplies = (ArrayList) entity.getProperty("replies");
-        BlogMessage message = new BlogMessage(messageId, tag, comment, image, nickname, email, messageReplies, timestamp);
+        //String image = (String) entity.getProperty("imgUrl");
+        ArrayList<String> message_Replies = (ArrayList) entity.getProperty("replies");
```
martiKallas: message_Replies should be messageReplies to follow the style guide.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -76,97 +70,88 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-
-      String sender = getParameter(request, "sender", "Steven");
+      String sender = request.getParameter("sender");
+      String loadFactor = request.getParameter("load_factor");
+      String postTag = getTagParameter(request, "tags", InternalTags.defaultTag());
       
-      // TODO: 
-      //      Get default tag from the InternalTags class and use that below.
+      //TODO: Handle image file sent with FormData.
+      //String imageUrl = getUploadedFileUrl(request, imagePath);
```
peizhaoou: Please don't comment out code. It's fine when debuggging, but please remember to double check every time before you send a PR.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -76,97 +70,88 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       
-      // Get type of comment.
-      String commentType = getParameter(request, "tags", "Default");
-
-      //String imageUrl = getUploadedFileUrl(request, "image");
+      // Only put BlogMessages with a message and tag in datastore.
+      if (message.equals("") || message.equals(null) || postTag == null) {
```
peizhaoou: For, message.equals(""), use message.isEmpty()

For message.equals(null), I believe it always returns false.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -76,97 +70,88 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       blogMessageEntity.setProperty("replies", messageReplies);
+    
       DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
       datastore.put(blogMessageEntity);
 
-      response.sendRedirect("/index.html");
+      // To get the recently posted message, add its tag to the |tagsToSearch| list.
+      // If the list has tags already, clear them before adding the tag.
+      // Our goal is to get the recently posted message.
+      if (!tagsToSearch.isEmpty()) {
```
peizhaoou: For lines 126-128, consider just call tagsToSearch.clear() without the conditional check

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -76,97 +70,88 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+        tagsToSearch.clear();
+      }
+      tagsToSearch.add(postTag);
+
+      // Everytime a post comes through, the javascript function that handles the Post assigns
+      // a load_factor value of 1.
+      // We assign that to the |numberOfCommentsToDisplay|. 
+      if (loadFactor != null) {
+        // we don't catch since this is always 1.
+        numberOfCommentsToDisplay = Integer.parseInt(loadFactor);
```
peizhaoou: Should we handle the potential parse exception here?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -76,97 +70,88 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      doGet(request, response);
     }
 
-  /**
-   * @return the request parameter, or the default value if the parameter
-   *         was not specified by the client.
-   */
-    private String getParameter(HttpServletRequest request, String name, String defaultValue) {
+    // Returns the requested parameter, or the default value if the parameter is null.
+    private String getTagParameter(HttpServletRequest request, String name, String defaultValue) {
```
peizhaoou: Why do we need this change instead of reusing getParameter()? getParameter() sounds more general.

portfolio/src/main/webapp/script.js
```diff
@@ -34,39 +34,101 @@ function openPage(pageName, elmnt, color) {
...
-    statsListElement.innerHTML = '';
+// Posts and fetches messages from the server and adds them to the DOM.
+async function sendAndGet() {
+  // Get post parameters.
+  const textEle = document.getElementById('text-input');
+  const nameEle = document.getElementById('sender');
+  const tagEle = document.getElementById('tags');
+
+  // To load the recent post BlogHashMap needs a loadFactor(limit) of 1.
+  const loadFactor = 1;
```
peizhaoou: what is the load factor?

portfolio/src/main/webapp/script.js
```diff
@@ -34,39 +34,101 @@ function openPage(pageName, elmnt, color) {
...
+
+  //TODO: use FormData function to post image file.
```
peizhaoou: Nit: fix comment style

portfolio/src/main/webapp/script.js
```diff
@@ -34,39 +34,101 @@ function openPage(pageName, elmnt, color) {
...
+  // const imgEle = document.getElementById('image');
```
peizhaoou: Remove commented-out code

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -71,108 +72,86 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      // Our goal is to get the recently posted message.
+      if (!tagsToSearch.isEmpty()) {
+        tagsToSearch.clear();
+      }
+      tagsToSearch.add(postTag);
+
+      // Everytime a post comes through, the javascript function that handles the Post assigns
+      // a load_factor value of 1.
+      // We assign that to the |numberOfCommentsToDisplay|. 
+      if (loadFactor != null) {
```
martiKallas: I don't really understand this section. You say that loadFactor will always be 1, so why shouldn't we just set numberOfCommentsToDisplay to 1?

portfolio/src/main/webapp/script.js
```diff
@@ -34,39 +34,101 @@ function openPage(pageName, elmnt, color) {
...
+async function loadPosts(){
+  fetch('/data').then(response => response.json()).then((msgs) => {
+    const statsListElement = document.getElementById('posts-list');
     msgs.forEach((msg) => {
-        statsListElement.appendChild(
-            createListElement(msg.sender + ': ' + msg.message));
-        statsListElement.appendChild(
-            createImgElement(msg.imgUrl));
+      statsListElement.appendChild(createListElement(msg));
+      //statsListElement.appendChild(createImgElement(msg.imgUrl));
```
peizhaoou: Same here.

portfolio/src/main/webapp/script.js
```diff
@@ -34,39 +34,101 @@ function openPage(pageName, elmnt, color) {
...
+  params.append('sender', nameEle.value);
+  params.append('tags', tagEle.value);
+  params.append('load_factor', loadFactor);
+
+  // Post parameters to the server and fetch instantly to build the page.
+  fetch('/data', {method: 'POST', body: params}).then(response => response.json()).then((msgs) => {
+    const statsListElement = document.getElementById('posts-list');
+    msgs.forEach((msg) => {
+      statsListElement.appendChild(createListElement(msg));
+      //statsListElement.appendChild(createImgElement(msg.imgUrl));
```
peizhaoou: Same here.

portfolio/src/main/webapp/script.js
```diff
@@ -38,54 +38,101 @@ function defaultPage() {
...
-    const statsListElement = document.getElementById('comments-container');
-    statsListElement.innerHTML = '';
+// Posts and fetches messages from the server and adds them to the DOM.
+async function sendAndGet() {
+  // Get post parameters.
+  const textEle = document.getElementById('text-input');
+  const nameEle = document.getElementById('sender');
+  const tagEle = document.getElementById('tags');
+
+  // To load the recent post BlogHashMap needs a loadFactor(limit) of 1.
```
martiKallas: If I understand this correctly, loadFactor isn't a good variable name. You use loadFactor in the servlet to change numberOfComments, so the variable name should reflect that. Also, load factor has a specific meaning for hash maps/tables, and so you should only use that if you are change the load factor of the hash map.

Load Factor for a hash map is the maximum number of occupied spots to empty spots before a hash map increases its size. If your load factor is 0.5, then a hash map will increase its size just after 50% of its spaces are filled.

portfolio/src/main/webapp/script.js
```diff
@@ -34,39 +34,101 @@ function openPage(pageName, elmnt, color) {
...
   imgElement.src = text;
   return imgElement;
 }
  
-function deleteData(){
-  fetch('/delete-data', {method: 'POST'}).then(getComments());
+async function deleteMessage(msg) {
+  const params = new URLSearchParams();
+  params.append('messageId', msg.id);
+  fetch('/delete-data', {method: 'POST', body: params}).then(response => response.text()).then((text) => {
```
peizhaoou: Line too long, please break (80 characters per line for JS)

portfolio/src/main/webapp/index.html
```diff
@@ -13,44 +13,42 @@
...
-        <form action="/delete-data" method="POST">
-          <button onclick="deleteData()">Delete Comments</button>
+        <form id="my-form" enctype="multipart/form-data">
+          <textarea id="sender" rows="1" cols="20" placeholder="Nickname"></textarea> 
+          <br>
+          <textarea id="text-input" rows="6" cols="60" placeholder="What's on your mind..."></textarea>
+          <br>
+          <label for="tags">Choose tags:</label>
+          <select name="tags" id="tags">
+            <option value="#general">Default</option>
```
peizhaoou, line 43: Hmm, why would we hard code these tags on the client side? I thought they should be something sent from the server side on loading the pages.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -71,108 +72,86 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         String comment = (String) entity.getProperty("text");
         String nickname = (String) entity.getProperty("nickname");
         String email = (String) userService.getCurrentUser().getEmail();
-        String image = (String) entity.getProperty("imgUrl");
-        ArrayList<String> messageReplies = (ArrayList) entity.getProperty("replies");
-        BlogMessage message = new BlogMessage(messageId, tag, comment, image, nickname, email, messageReplies, timestamp);
+        //String image = (String) entity.getProperty("imgUrl");
```
martiKallas: Commented out code shouldn't be checked into production. The main reasons are that it makes things less readable and it is easily forgotten about.

To preserve code like this, you can start a branch that you intend to work on later, or you can copy and paste from github's history.

portfolio/src/main/webapp/index.html
peizhaoou: Actually, never mind.

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -22,17 +22,17 @@
   private final String message;
   private final String nickname;
   private final long timestamp;
-  private final String image;
+  //private final String image;
```
peizhaoou: Nit: remove

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -51,9 +51,9 @@ public String getMessage() {
     return message;
   }
 
-  public String getImgUrl() {
-    return image;
-  }
+//   public String getImgUrl() {
```
peizhaoou: Same here.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -87,11 +87,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     
       Gson gson = new Gson();
       response.setContentType("application/json;");
-
-      List<BlogMessage> limitedBlogMessages = new ArrayList<>();
-      for (int i = 0; i < numberOfCommentsToDisplay; i++) {
-        limitedBlogMessages.add(allBlogMessages.get(i));
-      }
+      
       response.getWriter().println(gson.toJson(limitedBlogMessages));
```
peizhaoou: Hmm I am confused about what "limitedBlogMessages" is. why do we need to remove 90-94?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -116,7 +112,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       Entity blogMessageEntity = new Entity("blogMessage");
       blogMessageEntity.setProperty("nickname", nickname);
       blogMessageEntity.setProperty("text", message);
-      blogMessageEntity.setProperty("imgUrl", imageUrl);
+      // blogMessageEntity.setProperty("imgUrl", imageUrl);
```
peizhaoou: Remove?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -47,18 +48,18 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Iterator;
 
-/** Servlet that returns some example content. TODO: modify this file to handle comments data */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
     int numberOfCommentsToDisplay = 0;
```
martiKallas: I know this isn't due to your change, but could you fix the indents here?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: oh i meant to say "will always be a number". The previous implementation had 2 post functionalities, each of which required a different numberOfCommentsToDisplay value. so the idea was to assign the value that each post function required to the numberOfCommentsToDisplay.
But when i took away one of the post functions i forgot to make the change here to simply set numberOfCommentsToDisplay to 1.
So you're right I can just set it to 1 now.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: I thought since this value will always be a number, we didn't need to catch?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: In our case, this function is only useful for the getting the tag because if a user fails to include a tag we want to assign the defautTag. We can't assign default values for a user's name or message.
_**I should have changed the parameter names to sth like (...request, String Tag, String defaultTag) to make it's functionality more obvious.**_

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: Umm I never used "limitedBlogMessages". I think it happened when I tried to merge master into my branch to resolve conflicts after i made the PR. When I took 90-94 out i forgot to also change that part to "loadedBlogMessages".

Andrewmob2703-zz: But we don't need those lines because the BlogHashMap sends the requested amount of messages.
Those lines were written before the BlogHashMap implementation and were already in master. So when i merged master into my branch, things got mixed up

peizhaoou: So when you merge and resolve, you would normally either try make a judgement call yourself and remove it, or ask the original author about what is going on, confirm with them if you should just remove it. In the end of the day, this is a shared codebase (everyone has to maintain it).

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
peizhaoou: My understanding is this is something passed from the client side, which means we can't assume they satisfy the precondition you mentioned?

Andrewmob2703-zz:
> actually the same question to "tagsToSearch" & "messageReplies"



> actually the same question to "tagsToSearch" & "messageReplies"

Because these variable are sort of dynamic throughout the class. We plan to use multiple POST and GET js functions which will alter the values of these variables throughout the class. We need them to be like global variables

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -52,14 +52,25 @@
...
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
-  int numberOfCommentsToDisplay = 0;
-  List<String> tagsToSearch = new ArrayList<String>();
-  List<String> messageReplies = new ArrayList<String>(); 
-  // TODO: Handle replies later.
+  private int numberOfCommentsToDisplay = 0;
+  private List<String> tagsToSearch = new ArrayList<String>();
+  
+  private void putBlogsInDatastore(String tag, String message, String nickname, reply) {
```
peizhaoou: Does this compile? "reply" doesn't have a type.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -74,52 +85,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       String nickname = (String) entity.getProperty("nickname");
       String email = (String) userService.getCurrentUser().getEmail();
       ArrayList<String> messageReplies = (ArrayList) entity.getProperty("replies");
-      BlogMessage message = new BlogMessage(messageId, tag, comment, nickname, email, messageReplies, timestamp);
+      BlogMessage message = new BlogMessage(
+            messageId, tag, comment, nickname, email, messageReplies, timestamp);
       messages.add(message);
```
peizhaoou: Again does this compile at all? I don't know where we define "messages", do you mean "BlogMessages"?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -52,14 +52,25 @@
...
     
-  @Override
-  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException { 
-    List<BlogMessage> messages = new ArrayList<>();
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(blogMessageEntity);
+  }
+
+  private List<BlogMessage> getBlogsFromDatastore() {
+    List<BlogMessage> BlogMessages = new ArrayList<>();
```
peizhaoou: Fix variable naming.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
peizhaoou: If this doesn't compile, please always ensure the code compiles before a commit. The code compiles (and actually runs) is one key assumption in the code base and also for reviewers.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -74,52 +85,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    LinkedList<BlogMessage> loadedBlogMessages = blogMap.getMessages(tagsToSearch, numberOfCommentsToDisplay);
-    
+    LinkedList<BlogMessage> loadedBlogMessages = blogMap.getMessages(
+        tagsToSearch, numberOfCommentsToDisplay);
+
     Gson gson = new Gson();
     response.setContentType("application/json;");
       
     response.getWriter().println(gson.toJson(loadedBlogMessages));
     return;
```
peizhaoou, line 113: Remove the redundant "return"?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -74,52 +85,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    }
+  }
     
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get Post parameters.
-    String message = request.getParameter("text-input");
-    String nickname = request.getParameter("sender");
+    final String MESSAGE = request.getParameter("text-input");
```
peizhaoou: When I said line 118&119 should be constants, I actually meant the parameter "text-input" & "sender". The variable themselves can be just local variables like before.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -74,52 +85,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    final String NICKNAME = request.getParameter("sender");
     String postTag = request.getParameter("tags");
```
peizhaoou: Same for "tags".

peizhaoou:
> Because these variable are sort of dynamic throughout the class. We plan to use multiple POST and GET js functions which will alter the values of these variables throughout the class. We need them to be like global variables

Fair enough. In general, try to be cautious about using class variable if you could use local variables because local variables give you a narrow point of focus to a specific scope, where for "global" variables, you would need to see where else do we modify the value, etc.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: I'm not entirely sure, if response doesn't end the function then i don't think the return is redundant.

peizhaoou (approved):
LGTM overall. Please feel free to merge once you address the new comments.

PR #34, opened 2020-07-20
Steven curated posts

smarquez-jpg:
Built framework where curated posts will populate

### DIFF
--- a/portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
@@ -58,7 +58,6 @@ public class FollowedTagsServlet extends HttpServlet {
 
     }
     
-
     /**
    * Converts a ServerStats instance into a JSON string using the Gson library
    */
@@ -66,11 +65,13 @@ public class FollowedTagsServlet extends HttpServlet {
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
       UserService userService = UserServiceFactory.getUserService();
      
-      String commentType = getParameter(request, "tags", "#general");
+      String commentType = request.getParameter("tags");
+      if (commentType == null) {
+        return;
+      }
       String email = (String) userService.getCurrentUser().getEmail();
 
       followedPosts.add(commentType);
-      System.out.println(commentType);
 
       Entity followedTag = new Entity("followedTag");
       followedTag.setProperty("tag", commentType);
@@ -83,11 +84,4 @@ public class FollowedTagsServlet extends HttpServlet {
    * @return the request parameter, or the default value if the parameter
    *         was not specified by the client.
    */
-   private String getParameter(HttpServletRequest request, String name, String defaultValue) {
-      String value = request.getParameter(name);
-      if (value == null) {
-        return defaultValue;
-      }
-      return value;
-    }
 }
--- a/portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
@@ -1,0 +1,93 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.blobstore.BlobInfo;
+import com.google.appengine.api.blobstore.BlobInfoFactory;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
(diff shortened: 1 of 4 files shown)


peizhaoou (approved):

peizhaoou (commented):
Is this commit "Changed how info is sent to POST of followedtagsservlet and impliment..." a separate change compared to the commits in the previous PR? If so, please send a separate PR; othewise, its starts to get confusing as to whether you are fixing previous comments or start to add new changes.

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    
+
+    /**
+   * Converts a ServerStats instance into a JSON string using the Gson library
+   */
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      UserService userService = UserServiceFactory.getUserService();
+     
+      String commentType = getParameter(request, "tags", "#general");
```
Andrewmob2703-zz: i think we don't need the getParameter function since a user can decide to follow tags or not. If they choose not to we don't want to assign a default tag, because then we'll record that they follow the default tag even when they actually dont.

Andrewmob2703-zz: If they don't follow any tag we can just return so that nothing is saved in datastore

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+  /**
+   * @return the request parameter, or the default value if the parameter
+   *         was not specified by the client.
+   */
+   private String getParameter(HttpServletRequest request, String name, String defaultValue) {
+      String value = request.getParameter(name);
+      if (value == null) {
+        return defaultValue;
```
Andrewmob2703-zz: Same comment about not needing a default value here

portfolio/src/main/webapp/script.js
```diff
@@ -112,3 +112,12 @@ function showPostForm() {
 function setNicknameForm() {
     window.open("/login");
 }
+
+function sendFollowedTags() {
+  const tagElem = document.getElementById('followtags');
+
+  const params = new URLSearchParams();
+  params.append('tags', tagElem.value);
```
Andrewmob2703-zz, line 193: I think this will cause an error. tagElem has multiple values so you can loop through the options to access each value and check whether it was selected. (ie. for (var i = 0; i < tagElem.options.length; i++) {...}
And then you can access each option as tagElem.options[i]

Andrewmob2703-zz: Probably the easiest way is to directly access each option with their id's since you have id's for each option in the index file. Then you can check which ones were selected if any.

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
```diff
@@ -0,0 +1,93 @@
...
+@WebServlet("/follow-tags")
+public class FollowedTagsServlet extends HttpServlet {
+    List<String> followedPosts= new ArrayList<>();
+    
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    }
+    
+
```
martiKallas: nit: un-needed line here

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
martiKallas: QQ about this - how are you going to differentiate between a user who is following no tags and user who hasn't selected the tags they want to follow yet? Or a user who's request to follow tags go lost?

martiKallas (commented):
Sorry for the late review, I think I missed the notification email. Feel free to ping me when you submit a PR. Just address Andrew's comments and it LGTM.

portfolio/src/main/webapp/script.js
smarquez-jpg: I tested it out and from what I can tell it would appear that value of the tagElem is the value of the option selected so that way I wouldn't need to iterate through the list but that also means that the user can only pick one at a time and has to re submit if they want to follow multiple tags.

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
smarquez-jpg: I'm not sure how to handle a lost request but as far as the back end is concerned, right now there will either be a one or more instance of a user who has a tag that they follow in the datastore or there will be no instance of that user. But a user can exist outside of the datastore.

portfolio/src/main/webapp/script.js
Andrewmob2703-zz: ok. we can use that for now and try to improve it later

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
```diff
@@ -83,11 +84,4 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
    * @return the request parameter, or the default value if the parameter
    *         was not specified by the client.
    */
```
Andrewmob2703-zz, line 86: remove comment too?

Andrewmob2703-zz (approved):
LGTM. 
There's many imports. If we don't need all please consider keeping only what's needed.

portfolio/src/main/java/com/google/sps/servlets/FollowedTagsServlet.java
martiKallas: For lost request - this is just if someone responds with the tags they want to follow but the request gets lost on the way to the backend. Essentially like that request never happened.

This isn't something you need to solve in this PR, but I think the plan was to have a popup/redirect to ask users to put in their tag preference. If a user doesn't put in their tag preference, do you want the popup to show every time they visit the site? If not, you will need a way to track the fact that a user doesn't want to follow any tags.

PR #38, opened 2020-07-22
factor out doPost for testing

Andrewmob2703-zz:


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -186,4 +210,23 @@ private String getUploadedFileUrl(HttpServletRequest request, String formInputEl
       }
     }
 
+    // This mimics the doPost method above and is only for testing purposes.
+    public void doPost_ForTests(String msg, String nickname, String tag) {
```
martiKallas, line 192: This shouldn't be here. If you are going to make utility functions, they should be in their own file. Further, this should be called by doPost and it should be tested before it is submitted.

martiKallas (changes requested):
Can you please break this up into multiple PRs? It should be one that implements LoadFollowedTags, one that links LoadFollowedTags into the DataServlet, and one that handles the frontend changes.

Probably the easiest way to do this is to start new branches for those and copy and paste your changes.

PR #43, opened 2020-07-23
remove member variables and add more helper functions to improve func…

Andrewmob2703-zz:
…tionality.
doPost handles post requests and responds with the recent post.
doGet handles get requests and responds with posts using requested parameters.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -11,9 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
+ 
 package com.google.sps.servlets;
-
+ 
 import com.google.appengine.api.blobstore.BlobInfo;
 import com.google.appengine.api.blobstore.BlobInfoFactory;
 import com.google.appengine.api.blobstore.BlobstoreService;
@@ -31,8 +31,8 @@ import com.google.appengine.api.images.ServingUrlOptions;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.gson.Gson;
-import com.google.sps.data.BlogMessage;
 import com.google.sps.data.BlogHashMap;
+import com.google.sps.data.BlogMessage;
 import com.google.sps.data.InternalTags;
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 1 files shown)


smarquez-jpg (approved):
LGTM

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
-  }
-    
+ 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get BlogMessages from Datastore.
-    List<BlogMessage> BlogMessages = getBlogsFromDatastore();
-
-    // Separate posts from replies from BlogMessages. 
+    List<BlogMessage> blogMessages = LoadAllBlogsOrLast(true);
```
peizhaoou, line 64: Nit: usually you would put a nested comment for constant parameter to make it easier to read like:

LoadAllBlogsOrLast(/*all=*/true)

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+    // Separate posts from replies from |blogMessages|.
     // Add replies to messageReplies for the respective posts.
-    // Create BlogHashMap Object and put BlogMessages in the map.
     List<BlogMessage> blogMessagesReplies = new ArrayList<BlogMessage>();
-    for (BlogMessage message : BlogMessages) {
+    for (BlogMessage message : blogMessages) {
       if (message.getParentID() != 0) {
         blogMessagesReplies.add(message);
-        BlogMessages.remove(message);
+        blogMessages.remove(message);
```
peizhaoou, line 78: You probably cannot remove an element during iterating over the same container. Some reference[1]. It may mess up with the underlying iterator and throw exceptions.

[1] https://stackoverflow.com/questions/10431981/remove-elements-from-collection-while-iterating

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+ 
+    // Respond with the recent post.
+    // |LoadAllBlogsOrLast| returns the recent post if false is passed.
+    Gson gson = new Gson();
+ 
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(LoadAllBlogsOrLast(false)));   
+  }
+ 
+  // HELPER FUNCTION:.........................................................................
```
peizhaoou: The comment line may not be necessary? In some sense, all non-public methods are helper methods.

peizhaoou: It would be more useful to give some high-level description about what this method does, as well as description on non-trivial arguments.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+  private void putBlogsInDatastore(
+        String tag, String message, String nickname, List<BlogMessage> reply, long parentID) {
     // Only put BlogMessages with a message in datastore.
-    if (message == null || message.isEmpty()) {
-      return;
+    if (message != null || !message.isEmpty()) {
```
peizhaoou, line 151: A couple things:
(1) the condition checked is not correct, e.g., if message is null, it fails the first condition and tries to check the second one, giving an exception. I  think you are trying to do if (message != null && !message.isEmpty()) {...}.

(2) Try to check the opposite and return early so that the core logic is not nested.
if (message == null || message.isEmpty()) {
  return;
}
...

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
-
-      // To support running in Google Cloud Shell with AppEngine's dev server, we must use the relative
-      // path to the image, rather than the path returned by imagesService which contains a host.
-      try {
-        URL url = new URL(imagesService.getServingUrl(options));
-        return url.getPath();
-      } catch (MalformedURLException e) {
-        return imagesService.getServingUrl(options);
+ 
+  // HELPER FUNCTION:................................................................
```
peizhaoou: Same here. Maybe just remove line 163.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+  // Loads all BlogMessages from Datastore if true is passed,
+  // Otherwise if false is passed only the recent post is loaded.
+  // This is useful because each time a user posts, we only load the last BlogMessage
+  // Which reduces the time to load datastore.
+  private List<BlogMessage> LoadAllBlogsOrLast(boolean all) {
```
peizhaoou, line 169: To some degree, I think we should even further break apart this method into loadAllBlogs() & loadMostRecentBlogs(). This right now mix the two with an "all", which makes the underlying logic branched with a bool. If you really want to reuse the part (line 183-194), maybe add another getBlogMessageFromEntity().

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+        break;
       }
     }
-
-}
\ No newline at end of file
+    return blogMessages;
+  }
+ 
+  // HELPER FUNCTION:........................................................................
```
peizhaoou: Same here.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+  // Puts BlogMessages in BlogHashMap and loads the requested parameters.
+  private LinkedList<BlogMessage> sortAndLoadFromBlogHashMap(
```
peizhaoou, line 206: Try to give non-trivial arguments a description as well?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
peizhaoou: Elsewhere too

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+  }
+ 
+  // HELPER FUNCTION:........................................................................
+  // Returns a URL that points to the uploaded file, or null if the user didn't upload a file.
+  private String getUploadedFileUrl(HttpServletRequest request, String formInputElementName) {
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    Map<String, List<BlobKey>> blobs = blobstoreService.getUploads(request);
+    List<BlobKey> blobKeys = blobs.get(formInputElementName);
+ 
+    if (blobKeys == null || blobKeys.isEmpty()) {
```
peizhaoou, line 220: Move this check to the very beginning, even before you define those local variables.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
+    // To support running in Google Cloud Shell with AppEngine's dev server, we must use the relative
+    // path to the image, rather than the path returned by imagesService which contains a host.
+    try {
+      URL url = new URL(imagesService.getServingUrl(options));
+      return url.getPath();
+    } catch (MalformedURLException e) {
+      return imagesService.getServingUrl(options);
+    }
+  }
+
```
peizhaoou, line 250: Remove line?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,168 +49,203 @@
...
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
-  private int numberOfCommentsToDisplay = 0;
-  private List<String> tagsToSearch = new ArrayList<String>();
   private final static String MESSAGE_PARAMETER = "text-input";
   private final static String SENDER_PARAMETER = "sender";
   private final static String TAG_PARAMETER = "tags";
-  
-  // TODO: ADD TO FORM IN INDEX
-  private final static long PARENT_ID_PARAMETER = 0;
```
tayyaba17, line 62: I would ask not to delete this because I am going to update to actually get form input so then I'll have to go find where I placed the  PARENT_ID_PARAMETER

Andrewmob2703-zz: My bad i should've asked. So is it like a parameter name like the ones above it? I saw that you assigned 0 to it so i made it a local variable instead. We can meet to resolve conflicts.

martiKallas (commented):
Can you confirm that the only changes are removing the member variables, moving helper functions and including lines 89 - 101 from Steven's changes?

Andrewmob2703-zz (commented):
> Can you confirm that the only changes are removing the member variables, moving helper functions and including lines 89 - 101 from Steven's changes?

Yeah and now doPost doesn't call doGet. doPost handles post requests and responds with the recent post.
doGet handles get requests and responds with posts using requested parameters.
It's important for each to handle its own requests because some get requests (for instance, a user loading x amount of messages for x,y,z.. tags) change the load parameters. Post requests (ie. loading only recent post) also change the load parameters.

martiKallas:
> > Can you confirm that the only changes are removing the member variables, moving helper functions and including lines 89 - 101 from Steven's changes?
> 
> Yeah and now doPost doesn't call doGet. doPost handles post requests and responds with the recent post.
> doGet handles get requests and responds with posts using requested parameters.
> It's important for each to handle its own requests because some get requests (for instance, a user loading x amount of messages for x,y,z.. tags) change the load parameters. Post requests (ie. loading only recent post) also change the load parameters.

How does this affect the frontend? If I POST a message, does the message returned from doPost get added to the list of messages I see? This isn't something that needs to be addressed in this PR, but I'm curious how that will work.

Andrewmob2703-zz:
> > > Can you confirm that the only changes are removing the member variables, moving helper functions and including lines 89 - 101 from Steven's changes?
> > 
> > 
> > Yeah and now doPost doesn't call doGet. doPost handles post requests and responds with the recent post.
> > doGet handles get requests and responds with posts using requested parameters.
> > It's important for each to handle its own requests because some get requests (for instance, a user loading x amount of messages for x,y,z.. tags) change the load parameters. Post requests (ie. loading only recent post) also change the load parameters.
> 
> How does this affect the frontend? If I POST a message, does the message returned from doPost get added to the list of messages I see? This isn't something that needs to be addressed in this PR, but I'm curious how that will work.

Yes it gets appended to the whatever was there before you posted

martiKallas (approved):
LGTM. Just address Peizhao's comments and make sure you manually test this before submitting.

PR #46, opened 2020-07-23
add method to load posts on the home page

Andrewmob2703-zz:
only top 5 posts are displayed on the homepage.
Posts are non-interactive. ie. display-only

### DIFF
--- a/portfolio/src/main/webapp/script.js
+++ b/portfolio/src/main/webapp/script.js
@@ -66,14 +66,14 @@ async function sendAndGet() {
 // Makes a Get request and loads Posts to the Blog page on body load
 // Also loads 5 display-only posts to be displayed on the home page.
 async function loadPosts(){
+  const HOME_LOAD_AMOUNT = 5;
   fetch('/data').then(response => response.json()).then((msgs) => {
     const statsListElement = document.getElementById('posts-list');
     const homeListElement = document.getElementById('home-comments-container');
     var i = 1;
     msgs.forEach((msg) => {
       statsListElement.appendChild(createListElement(msg));
-      console.log('value :' + i);
-      if (i > 5) {
+      if (i > HOME_LOAD_AMOUNT) {
         return true;
       }
       homeListElement.appendChild(createListElementHome(msg));
@@ -100,7 +100,7 @@ function getCommentsTag(tag) {
 // Creates an <li> element containing message details.
 function createListElement(msg) {
   const postElement = document.createElement('li');
-  postElement.style.margin = "10px";
+  postElement.className = 'post';
 
   const messageElement = document.createElement('span');
   messageElement.innerText = msg.message;
@@ -113,6 +113,7 @@ function createListElement(msg) {
   }
   userElement.style.marginLeft = "15px";
 
+  // TODO: move styles to style.css
   const timeElement = document.createElement('span');
   var date = new Date(msg.timestamp);
   timeElement.innerText = date.toString().slice(0, 24);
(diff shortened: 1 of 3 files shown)


portfolio/src/main/webapp/script.js
```diff
@@ -63,24 +63,21 @@ async function sendAndGet() {
...
+// Makes a Get request and loads Posts to the Blog page on body load
+// Also loads 5 display-only posts to be displayed on the home page.
 async function loadPosts(){
   fetch('/data').then(response => response.json()).then((msgs) => {
     const statsListElement = document.getElementById('posts-list');
+    const homeListElement = document.getElementById('home-comments-container');
+    var i = 1;
     msgs.forEach((msg) => {
       statsListElement.appendChild(createListElement(msg));
+      console.log('value :' + i);
```
Andrewmob2703-zz: Meant to take out

portfolio/src/main/webapp/script.js
```diff
@@ -63,24 +63,21 @@ async function sendAndGet() {
...
+      if (i > 5) {
+        return true;
```
peizhaoou, line 77: What does this function return? I ask because this is the only place it returns a value (true).

portfolio/src/main/webapp/script.js
```diff
@@ -63,24 +63,21 @@ async function sendAndGet() {
...
+// Also loads 5 display-only posts to be displayed on the home page.
 async function loadPosts(){
   fetch('/data').then(response => response.json()).then((msgs) => {
     const statsListElement = document.getElementById('posts-list');
+    const homeListElement = document.getElementById('home-comments-container');
+    var i = 1;
     msgs.forEach((msg) => {
       statsListElement.appendChild(createListElement(msg));
+      console.log('value :' + i);
+      if (i > 5) {
```
peizhaoou: I'd recommend defining the "5" as a constant on the top of this file like:

const MAX_POST_NUMBER = 5;

Then use the variable here.

peizhaoou (approved):
Make sure you get another LGTM from the Pod as well.

portfolio/src/main/webapp/script.js
martiKallas: I would make this 5 a const instead of a 'magic number'. Maybe MAX_HOMEPAGE_POSTS or something.

portfolio/src/main/webapp/script.js
Andrewmob2703-zz: It doesn't return anything. I needed something like "continue", but in JQuery (in this case the forEach method) continue doesn't work. I did a quick search and it looks like returning true is how people do continue in jQuery.

https://stackoverflow.com/questions/7028716/illegal-continue-statement-in-javascript-each-loop

portfolio/src/main/webapp/script.js
```diff
@@ -140,6 +137,37 @@ function createListElement(msg) {
   return postElement;
 }
 
+// Creates a list of posts to be displayed on the homepage.
+// Display-only.(ie. No reply or delete button). 
+function createListElementHome(msg) {
+  const postElement = document.createElement('li');
+  postElement.style.margin = "10px";
```
martiKallas: It's probably better to do this in a css file and give your element a class to control the styling. That way your styling and processing logic are separate.

portfolio/src/main/webapp/script.js
```diff
@@ -140,6 +137,37 @@ function createListElement(msg) {
...
+  const messageElement = document.createElement('span');
+  messageElement.innerText = msg.message;
+  
+  const userElement = document.createElement('span');
+  if (msg.nickname === undefined || msg.nickname === null) {
+    userElement.innerHTML = "<b><i>_Anonymous</i></b>";
+  } else {
+    userElement.innerHTML = "<b><i>_" + msg.nickname + "</i></b>";
+  }
+  userElement.style.marginLeft = "15px";
```
martiKallas, line 162: Here too, try to keep style and processing logic separate.

portfolio/src/main/webapp/script.js
```diff
@@ -140,6 +137,37 @@ function createListElement(msg) {
...
+
+  const timeElement = document.createElement('span');
+  var date = new Date(msg.timestamp);
+  timeElement.innerText = date.toString().slice(0, 24);
+  timeElement.style.marginTop = "5px";
+  timeElement.style.float = "right";
+  timeElement.style.clear ="left";
```
martiKallas: Here too - keep styling separate.

martiKallas (approved):
Nice PR size, and I like how the 'theme' of this PR is consistent!

Just a few comments to address.

PR #48, opened 2020-07-24
Refactored doGet with function to put replies with posts outside of d…

smarquez-jpg:
…oGet

### DIFF
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -61,14 +61,11 @@ public class DataServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get BlogMessages from Datastore.
-    List<BlogMessage> blogMessages = LoadAllBlogsOrLast(/*all=*/ true);
-    
+    List<BlogMessage> blogMessages = putRepliesWithPosts(LoadAllBlogsOrLast(/*all=*/ true));
     // TODO: Get these from client.
     int numberOfCommentsToDisplay = 0;
     List<String> tagsToSearch = new ArrayList<String>();
  
-    blogMessages = putRepliesWithPosts(blogMessages);
-
     // Get user email
     UserService userService = UserServiceFactory.getUserService();
     String email = (String) userService.getCurrentUser().getEmail();
@@ -127,10 +124,6 @@ public class DataServlet extends HttpServlet {
   }
  
   private List<BlogMessage> putRepliesWithPosts(List<BlogMessage> blogMessages) {
-    
- 
-    // TODO: Use helper functions for lines 73-101.
- 
     // Separate posts from replies from |blogMessages|.
     // Add replies to messageReplies for the respective posts.
     List<BlogMessage> blogMessagesReplies = new ArrayList<BlogMessage>();


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -144,6 +126,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+ 
+    // TODO: Use helper functions for lines 73-101.
+ 
+    // Separate posts from replies from |blogMessages|.
+    // Add replies to messageReplies for the respective posts.
+    List<BlogMessage> blogMessagesReplies = new ArrayList<BlogMessage>();
+    for (BlogMessage message : blogMessages) {
+      if (message.getParentID() != 0) {
+        blogMessagesReplies.add(message);
+        blogMessages.remove(message);
```
Andrewmob2703-zz: Removing while iterating will cause a concurrent exception. But I think Tayyaba worked on it so she can change it later

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -144,6 +126,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     response.getWriter().println(gson.toJson(LoadAllBlogsOrLast(/*all=*/false)));   
   }
  
+  private List<BlogMessage> putRepliesWithPosts(List<BlogMessage> blogMessages) {
+    
+
```
Andrewmob2703-zz: Remove line?

Andrewmob2703-zz (approved):
LGTM! 
We just need to fix line 140 to avoid the exception

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -62,31 +62,13 @@
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get BlogMessages from Datastore.
     List<BlogMessage> blogMessages = LoadAllBlogsOrLast(/*all=*/ true);
- 
+
```
peizhaoou: Remove the empty space?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -62,31 +62,13 @@
...
- 
-    for (BlogMessage post : blogMessages) {
-      for (BlogMessage reply : blogMessagesReplies) {
-        if (reply.getParentID() == post.getTimestamp()) {
-          post.addReply(reply);
-        }
-      } 
-    }
- 
+    blogMessages = putRepliesWithPosts(blogMessages);
```
peizhaoou, line 68: I would combine line 64 and 70 into

List<BlogMessage> blogMessages = putRepliesWithPosts(LoadAllBlogsOrLast(/*all=*/true));

instead of overwriting the same variable.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -144,6 +126,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     response.getWriter().println(gson.toJson(LoadAllBlogsOrLast(/*all=*/false)));   
   }
  
+  private List<BlogMessage> putRepliesWithPosts(List<BlogMessage> blogMessages) {
+    
+ 
+    // TODO: Use helper functions for lines 73-101.
```
peizhaoou: I would generally avoid using very specific line numbers in TODOs because we will easily forget to update the line numbers when we change the code elsewhere, making us lose track of the numbers maybe in a day or two. Maybe try to use some slightly less specific description to refer to the range of code you mean.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -144,6 +126,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      if (message.getParentID() != 0) {
+        blogMessagesReplies.add(message);
+        blogMessages.remove(message);
+      }
+    }
+ 
+    for (BlogMessage post : blogMessages) {
+      for (BlogMessage reply : blogMessagesReplies) {
+        if (reply.getParentID() == post.getTimestamp()) {
+          post.addReply(reply);
```
peizhaoou: Similarly to Andrew's "removing while iterating" comment, you need to be very careful when you change the container when you iterate through it. Usually you would want to either avoid that or use explicitly iterator.

peizhaoou (dismissed):
Please address the comments about "modifying the container while iterating it".

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
Andrewmob2703-zz: +1. but in this case i think the container itself is not changing? post is not removed or added: one of its members, 'messageReplies' is getting another item 'reply'. so the iteration container doesn't change i think?

smarquez-jpg: Yeah, since reply is being added to the list of replies for the post it shouldn't change the number of items in blogMessages.

PR #49, opened 2020-07-24
Steven get blog messages

smarquez-jpg:


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -144,6 +126,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     response.getWriter().println(gson.toJson(LoadAllBlogsOrLast(/*all=*/false)));   
   }
  
+  private List<BlogMessage> putRepliesWithPosts(List<BlogMessage> blogMessages) {
+    
+
```
Andrewmob2703-zz: Remove line?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -144,6 +126,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+ 
+    // TODO: Use helper functions for lines 73-101.
+ 
+    // Separate posts from replies from |blogMessages|.
+    // Add replies to messageReplies for the respective posts.
+    List<BlogMessage> blogMessagesReplies = new ArrayList<BlogMessage>();
+    for (BlogMessage message : blogMessages) {
+      if (message.getParentID() != 0) {
+        blogMessagesReplies.add(message);
+        blogMessages.remove(message);
```
Andrewmob2703-zz, line 123: i think this was in the other PR. This will cause an exception but we can change that later when it starts to matter

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -151,6 +137,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     return blogMessages;
   }
 
+  private List<String> updateTagsToSearch (List<String> tagsToSearch) {
```
Andrewmob2703-zz: This function could probably have a better name like getUserFollowedTags.
Or some comments about what it does.

Andrewmob2703-zz (approved):
Looks good.
Few comments

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
peizhaoou: Another thing: why does this method need to return a List<String> instead of a "void"? I think it would be confusing what the return value is for. It is usually common to return void and directly update what is referenced by the argument.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -151,6 +137,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    UserService userService = UserServiceFactory.getUserService();
+    String email = (String) userService.getCurrentUser().getEmail();
+ 
+    // Check to see if user follows any tags
+    if (LoadFollowedTags.hasFollowedTags(email)) {
```
peizhaoou: Always think about returning early to avoid deep nesting like:

if (!LoadFollowedTags.hasFollowedTags(email)) {
  return;
}
// Your core logic.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -143,13 +140,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
  
     // Check to see if user follows any tags
     if (LoadFollowedTags.hasFollowedTags(email)) {
-      List<String> newTags = LoadFollowedTags.getFollowedTags(email);
-      for (String tag : newTags) {
-        if (!tagsToSearch.contains(tag)) {
-          tagsToSearch.add(tag);
-        }
-      } 
+      return tagsToSearch;
```
Andrewmob2703-zz: i think you want to return if user doesn't have followedTags. you missed the '!' in the condition

Andrewmob2703-zz (commented):
I think these 2 PRs are related and could be 1. Right now there's conflicts between them which can make merging a problem. You could copy your changes into one branch and commit inside 1 PR probably naming it "Put replies separation and userFollowedTags functionality in helper functions"?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
peizhaoou: In that case, you may want to return a bool to indicate whether you actually update the |tagsToSearch| argument or not. Remember to add a comment to explain what the return type is.

smarquez-jpg:
> Also update the return type to "void"?

I have it return a list right now because the function modifies tagsToSearch and since that's a local variable in the doGet I thought it would be best to have the function return the modified tagsToSearch since it still needs to be used in the response.getWriter() at the end.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -68,7 +68,6 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     List<String> tagsToSearch = new ArrayList<String>();
  
     blogMessages = putRepliesWithPosts(blogMessages);
-
     tagsToSearch = updateTagsToSearch(tagsToSearch);
```
peizhaoou: So the reason I said we don't need to return the "tagsToSearch" in updateTagsToSearch() method is that what you return right now is just the reference that we pass here in line 71. Say if you updated it to return "void", it does not affect here at all in the sense that you can continue to use variable |tagsToSearch|, e.g.:

 blogMessages = putRepliesWithPosts(blogMessages);
updateTagsToSearch(tagsToSearch);
...
response.getWriter().println(gson.toJson(sortAndLoadFromBlogHashMap(
    blogMessages, tagsToSearch, numberOfCommentsToDisplay)));
}

Note that passing |tagsToSearch| does not mean you lose the reference within the method. And updating it to "void" makes it less confusing because now readers know you take a reference and potentially update its content.

peizhaoou (approved):

PR #50, opened 2020-07-27
POST Reply

tayyaba17:
Replies are now succesfully posted to datastore! I have to hide some things from the user (the parentID div and tags-reply div). I also need to check out the DataServlet file before this can be merged because when there is a reply, there is an error that arises with loading that I am trying to debug. Also need to remove console.log() statements.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -55,6 +55,7 @@ public class DataServlet extends HttpServlet {
   private final static String MESSAGE_PARAMETER = "text-input";
   private final static String SENDER_PARAMETER = "sender";
   private final static String TAG_PARAMETER = "tags";
+  private final static String PARENTID_PARAMETER = "parentID";
   public final static String BLOG_ENTITY_KIND = "blogMessage";
   // Can add another parameter name for parentID which should be a constant.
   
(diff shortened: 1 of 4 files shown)


portfolio/src/main/webapp/script.js
```diff
@@ -39,19 +39,26 @@ function defaultPage() {
...
   params.append('text-input', textEle.value);
   params.append('sender', nameEle.value);
-  params.append('tags', tagEle.value);
+  params.append('parentID', parentidEle.value);
+
+  console.log("In SendAndGet");
+  console.log("This is the text value", textEle.value);
+  console.log("This is the sender value", nameEle.value);
+  console.log("This is the tag value", tagEle.value);
+  console.log("This is the parentID value", parentidEle.value);
```
tayyaba17: Will delete

smarquez-jpg: Don't forget to remove the print statements before merging into master

smarquez-jpg (approved):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -122,12 +123,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
  
     String nickname = request.getParameter(SENDER_PARAMETER);
     String postTag = request.getParameter(TAG_PARAMETER);
+    String parentIDString = request.getParameter(PARENTID_PARAMETER);
+    long parentID = Long.parseLong(parentIDString);
```
peizhaoou, line 94: You would probably want to handle the parse exception here if |parentIDString| is not a valid long integer.

peizhaoou: For better error messages to the client side.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -122,12 +123,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     if (postTag == null || postTag.isEmpty()) {
       postTag = InternalTags.defaultTag();
     }
     
     // TODO: Get parentID from client when a reply is made.
-    long parentID = 0;
+//    long parentID = 0;
```
peizhaoou: Nit: remove?

peizhaoou (approved):
Mostly LGTM

PR #52, opened 2020-07-28
Fix nullPointerException and concurrentModificationException

tayyaba17:


portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -34,7 +34,11 @@ public BlogMessage(long id, String tag, String message,
     this.message = message;
     this.nickname = nickname;
     this.email = email;
-    this.messageReplies = messageReplies;
+    if (messageReplies != null) {
+      this.messageReplies = messageReplies;   
+    } else {
+       this.messageReplies = new ArrayList<BlogMessage>();
```
Andrewmob2703-zz, line 40: I'm not sure if it's conventional to check one member for null in a constructor with many members or if it's clear to someone that doesn't know why you're doing this (**Marti/Peizhao thoughts?**) 
But to avoid this, and since messageReplies contain BlogMessages, we can define a default constructor with no params and use it for messageReplies when we're calling the constructor with params.

martiKallas: I'm not very familiar with java, but this should be OK. Just note in the comments for the constructor which parameters can be null.

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -59,8 +63,10 @@ public String getSender() {
     return messageReplies;
   }
 
-  public void addReply(BlogMessage message) {
-    messageReplies.add(message);
+  public void addReply(BlogMessage reply) {
```
Andrewmob2703-zz, line 66: Since this just adds a reply to messageReplies, i think it's not needed?
You could use the getReplies function to get the list and add to it. 
Like post.getReplies().add(reply).

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -111,25 +111,27 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    for (BlogMessage message : blogMessagesAll) {
+      if (message.getParentID() == 0) {
+        blogMessagesParents.add(message);
+      } else {
+          blogMessagesReplies.add(message);
       }
     }
  
-    for (BlogMessage post : blogMessages) {
+    for (BlogMessage post : blogMessagesParents) {
```
martiKallas, line 127: Is there a data structure you could here that would eliminate the need for a double for loop?

Andrewmob2703-zz (commented):
for the concurrentException fix, you can make a list of nonPosts and add messages that are not posts to it and then do blogMessagesAll.removeAll(nonPosts) afer iteration. That way you dont need to define the blogMsgsParents. And also you modify the list you take and return the modified list

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
Andrewmob2703-zz: ok

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
martiKallas: This would be correct if addReply were a utility function. However, it is usually good Object Oriented design to keep member variables private whenever possible. In this case using the 'setter' addReply is a good choice because it allows users of BlogMessage to add replies to a post, but it prevents them from doing whatever they want with messageReplies.

martiKallas (approved):
This is a very good PR size. Don't be afraid to go even smaller. You could have broken this into one PR for each file.  

You don't have to address the double for loop in this PR (in the name of incremental changes), but if you do, please ping your reviewers again so they can double check the work.

Andrewmob2703-zz (approved):
Looks good!
We can refactor later

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
tayyaba17: ooh.. i could probably use a hashtable and make the keys the id's making it more efficient.

PR #54, opened 2020-07-29
 Remove replies property from datastore

Andrewmob2703-zz:
Since replies are now stored as BlogMessages, the "replies" property in datastore will always be null, which is what was causing the nullPtr exception.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/BlogMessage.java
+++ b/portfolio/src/main/java/com/google/sps/data/BlogMessage.java
@@ -34,11 +34,7 @@ public final class BlogMessage {
     this.message = message;
     this.nickname = nickname;
     this.email = email;
-    if (messageReplies != null) {
-      this.messageReplies = messageReplies;   
-    } else {
-       this.messageReplies = new ArrayList<BlogMessage>();
-    }
+    this.messageReplies = messageReplies;   
     this.timestamp = timestamp;
     this.parentID = parentID;
   }
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -52,16 +52,16 @@ import java.util.Map;
  
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
-  // Can add another parameter name for parentID which should be a constant.
   
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Get BlogMessages from Datastore.
+    // Get BlogMessages from Datastore and put replies with their respective posts.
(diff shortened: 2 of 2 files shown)


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -52,16 +52,16 @@
...
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Get BlogMessages from Datastore.
+    // Get BlogMessages from Datastore and put replies with their respective posts.
     List<BlogMessage> blogMessages = putRepliesWithPosts(LoadAllBlogsOrLast(/*all=*/ true));
+
     // TODO: Get these from client.
     int numberOfCommentsToDisplay = 0;
     List<String> tagsToSearch = new ArrayList<String>();
-    blogMessages = putRepliesWithPosts(blogMessages);
```
Andrewmob2703-zz, line 64: This is done above.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -122,6 +119,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     for (BlogMessage post : blogMessagesParents) {
       for (BlogMessage reply : blogMessagesReplies) {
         if (reply.getParentID() == post.getMessageId()) {
+          System.out.println("we're adding a rep");
```
Andrewmob2703-zz, line 122: Will remove

smarquez-jpg (approved):
LGTM after the println is removed

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
```diff
@@ -34,11 +34,7 @@ public BlogMessage(long id, String tag, String message,
     this.message = message;
     this.nickname = nickname;
     this.email = email;
-    if (messageReplies != null) {
-      this.messageReplies = messageReplies;   
-    } else {
-       this.messageReplies = new ArrayList<BlogMessage>();
-    }
+    this.messageReplies = messageReplies;
```
peizhaoou, line 37: If I recall correctly, this is what Tayyaba added to check whether the passed-in messageReplies is null or not. Is this not needed anymore?

peizhaoou (commented):
Other than the comment in the constructor of BlogMessage, this LGTM. If you already synced with Tayyaba on this, feel free to merge.

portfolio/src/main/java/com/google/sps/data/BlogMessage.java
Andrewmob2703-zz: Yes the null was coming from the "replies" property in datastore which was always going to be null since replies are now stored as BlogMessages. So by removing that property from datastore we avoid the null problem

PR #57, opened 2020-07-30
Curated Posts Test

tayyaba17:
Here's a (hopefully) small PR for the testing.

### DIFF
--- a/portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
+++ b/portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
@@ -2,7 +2,6 @@ package com.google.sps.servlets;
 
 import static com.google.appengine.api.datastore.FetchOptions.Builder.withLimit;
 import static org.junit.Assert.assertEquals;
-
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
@@ -16,6 +15,11 @@ import java.util.ArrayList;
 import java.util.List;
 
 public class CuratedPostsTest {
+  public static String TAG = "tag";
+  public static String NO_TAGS_EMAIL = "noTags@testing.com";
+  public static String HAS_TAGS_EMAIL = "hasTags@testing.com";
+  public static String HAS_MULTIPLE_TAGS_EMAIL = "hasMultipleTags@testing.com";
+  public static DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
   private final LocalServiceTestHelper helper =
     new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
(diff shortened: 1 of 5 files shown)


portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,75 @@
...
+  @Test
+  public void testFollowsOneTag() {
+    createDatastoreEntities();
+    List<String> expectedFollowedTags = new ArrayList<String>();
+    expectedFollowedTags.add("#business");
+    assertEquals(expectedFollowedTags, LoadFollowedTags.getFollowedTags("hasTags@testing.com"));
+  }
+
+  @Test
+  public void testFollowsNoTags() {
```
smarquez-jpg: Looks good, but you might want to take this out until it's filled in before you merge.

smarquez-jpg (approved):

smarquez-jpg (approved):

Andrewmob2703-zz (commented):
Thanks for separating the 2. 
**_at LoginServlet 72_**
But you still want the **id next to the "UserLogin" like ("UserLogin", id) .**
That way if a user changes their nickname it's overrided in datastore and we don't store multiple nicknames for one user.
Also consider changing "UserLogin" to maybe "BlogUsers"?

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  private void createDatastoreEntities() {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    Entity noTagEntity = new Entity("UserLogin");
```
martiKallas: This is lower priority, but consider creating a private helper function to add entities to Datastore. You could even make this a util class that can be used throughout testing and production code. This code is replicated in many locations. 

You could have a function like storeUserLoginEntity(id, nickname, List<tags>), and others for the other Entities.

I don't know if it is expensive to call getDatastoreService, but if it is, you can pass the DatastoreService as an argument. If the Datastore documentation doesn't say anything about the call to getDatastoreService being expensive, then it shouldn't be.

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Test
+  public void testFollowsNoTags() {
+    createDatastoreEntities();
+    List<String> expectedFollowedTags = new ArrayList<String>();
+    assertEquals(expectedFollowedTags, LoadFollowedTags.getFollowedTags("noTags@testing.com"));
+  }
+
+  @Test
+  public void testFollowsOneTag() {
```
martiKallas, line 87: I like the thorough testing, but you probably don't need this function. Does this test a different path through the code than the testFollowsMultipleTag? It's not a bad thing, but it's something you can think about in the future to save yourself a little work.

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,122 @@
...
+    helper.tearDown();
+  }
+
+  private void createDatastoreEntities() {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    Entity noTagEntity = new Entity("UserLogin");
+    noTagEntity.setProperty("id", "357911");
+    noTagEntity.setProperty("nickname", "noTags");
+    noTagEntity.setProperty("email", "noTags@testing.com");
```
martiKallas: As much as possible, avoid bare strings for values that you use in multiple places. For any value that you compare against in your tests, you should have a constant for it. The fields that aren't compared are fine as bare strings. Using a constant creates a clear link between the setup for the test and how it is used in the actual testing.

You should also use the BlogConstants for the properties names. If they ever change, then a person only needs to change the one file without changing your test.

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,122 @@
...
+  @Test
+  public void testFollowsOneTag() {
+    createDatastoreEntities();
+    List<String> expectedFollowedTags = new ArrayList<String>();
+    expectedFollowedTags.add("#business");
+    assertEquals(expectedFollowedTags, LoadFollowedTags.getFollowedTags("hasTags@testing.com"));
+    assertEquals(1, LoadFollowedTags.getFollowedTags("hasTags@testing.com").size());
+  }
+
+    @Test
```
martiKallas: nit: indentation

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,122 @@
...
+    hasMultipleTagsFollowed.add("#education");
+    hasMultipleTagsFollowed.add("#music");
+    
+    for (int i=0; i<hasMultipleTagsFollowed.size(); i++) {
+      Entity hasMultipleTagsFollowedTags = new Entity("followedTag");
+      hasMultipleTagsFollowedTags.setProperty("tag", hasMultipleTagsFollowed.get(i));
+      hasMultipleTagsFollowedTags.setProperty("email", "hasMultipleTags@testing.com");
+      datastore.put(hasMultipleTagsFollowedTags);
+    }
+
```
martiKallas: nit: don't need this blank line

martiKallas (approved):
I left a few comments before you submit. I like the organization and structure of this change!

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import java.util.ArrayList;
+import java.util.List;
+
+
```
peizhaoou: Remove empty lines.

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class CuratedPostsTest {
+
```
peizhaoou, line 23: Remove extra line.

peizhaoou: Cool never mind you did that

portfolio/src/test/java/com/google/sps/CuratedPostsTest.java
tayyaba17: I was going off based on the doc. I think the idea may be to catch if curated posts is working for just one tag opposed to multiple?

PR #58, opened 2020-07-30
add functionality for reply box to appear under a post when its reply…

Andrewmob2703-zz:
… button is clicked.
Also,
sendAndGet is removed; Added submit event listeners on both forms to use Fetch to post to the server once a form's submit button is clicked.

### DIFF
--- a/portfolio/src/main/webapp/script.js
+++ b/portfolio/src/main/webapp/script.js
@@ -11,7 +11,65 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
- 
+
+
+// These variables will be used to track whenever a reply button is clicked
+// And whether the same button was clicked again in a row.
+var replyClickCounter = 0;
+var currentReplyBtnClicked = 0;
+
+// Post event listener to submit posts with Fetch on form submit. 
+document.getElementById('post-form').addEventListener('submit', function(postSubmitEvent) {
+  // Prevent form from redirecting to the server.
+  postSubmitEvent.preventDefault();
+  
+  // Get a reference to all the elements on the form.
+  const formData = new FormData(this);
+  const postFormParams = new URLSearchParams();
+  
+  // Append a pair [parameter name: value] of the form's elements
+  // to a URLSearchParams.
+  for (const pair of formData) {
+    postFormParams.append(pair[0], pair[1]);
+  }
+
+  // Make a POST request with the new post and fetch it to add to the DOM.
+  fetch('/data', {method: 'POST', body: postFormParams}).then(
+        response => response.json()).then((msgs) => {
+    const statsListElement = document.getElementById('posts-list');
+    msgs.forEach((msg) => {
+      statsListElement.appendChild(createListElement(msg));
+    })
+  }).catch (function(error) {
+      console.log(error);
+      // No form reset if post wasn't submitted; return.
+      return;
+  })
+  
+  // Empty the post form after a post is submitted.
+  resetForm('post-form');
+});
+
+// Reply event listener to submit replies with Fetch on form submit.
+document.getElementById('reply-form').addEventListener('submit', function(replySubmitEvent) {
+  replySubmitEvent.preventDefault();
+
+  const formData = new FormData(this);
+  const replyFormParams = new URLSearchParams();
+  
+  for (const pair of formData) {
+    replyFormParams.append(pair[0], pair[1]);
+  }
+
+  fetch('/data', {method: 'POST', body: replyFormParams}).catch (function(error) {
+      console.log(error);
+      return;
+  });
+  // TODO: display replies under their respective posts.
+
+  resetForm('reply-form');
+});
+
 function openPage(pageName, elmnt, color) {
   // Hide all elements with class="tabcontent" by default */
   var i, tabcontent, tablinks;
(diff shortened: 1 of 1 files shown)


tayyaba17:
Pretty big PR, please try to break it down into smaller pieces next time so it's easier to follow

tayyaba17:
LGTM

Andrewmob2703-zz:
> Pretty big PR, please try to break it down into smaller pieces next time so it's easier to follow

Noted

portfolio/src/main/webapp/script.js
```diff
@@ -11,7 +11,65 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
- 
+
+
+// These variables will be used to track whenever a reply button is clicked
+// And whether the same button was clicked again in a row.
+var replyClickCounter = 0;
```
peizhaoou, line 18: For the counter, I find it a bit confusing because the only reason you use it seems to be check whether the click counts clicked with an odd number of times. I think it would be more straightforward to use a boolean variable like a "replyClickToogle" or something to mark if it's enabled or disabled.

portfolio/src/main/webapp/script.js
```diff
@@ -11,7 +11,65 @@
...
+  
+  // Get a reference to all the elements on the form.
+  const formData = new FormData(this);
+  const postFormParams = new URLSearchParams();
+  
+  // Append a pair [parameter name: value] of the form's elements
+  // to a URLSearchParams.
+  for (const pair of formData) {
+    postFormParams.append(pair[0], pair[1]);
+  }
```
martiKallas, line 34: nit: This part is a little confusing because you are appending Form data to URLSearchParams, but then passing it in the body of the POST method which is where Form data would normally go. Typically, URLSearchParams are appended to the end of the URL, and are not passed as part of the body of the message.

I think it would make more sense to append the URLSearchParams to formData because you are ultimately sending the data the way Form data would normally be sent. I don't know if this is actually possible with these objects, but it looks like it should be with the code you have.

portfolio/src/main/webapp/script.js
```diff
@@ -11,7 +11,65 @@
...
+// Reply event listener to submit replies with Fetch on form submit.
+document.getElementById('reply-form').addEventListener('submit', function(replySubmitEvent) {
+  replySubmitEvent.preventDefault();
+
+  const formData = new FormData(this);
+  const replyFormParams = new URLSearchParams();
+  
+  for (const pair of formData) {
+    replyFormParams.append(pair[0], pair[1]);
+  }
```
martiKallas, line 62: Same point here as above... it would make more sense to append the URLSearchParams to the formData (if possible).

portfolio/src/main/webapp/script.js
```diff
@@ -199,16 +244,69 @@ async function deleteMessage(msg) {
...
+// Removes the reply form if button is clicked again, and so on...
+function handleReplies(msgID, msgSender, tag, postEleID) { 
+  // The form appears when a reply button is clicked an even number of times, so
+  // If a users clicks a different reply button, set counter to zero for that button
+  if (clickedDifferentReplyButton(msgID)) {
+    replyClickCounter = 0;
+  }
+  
+  // Remove the form when a button is clicked an odd number of times in a row.
+  if (replyClickCounter % 2 != 0) {
```
martiKallas, line 257: I think a cleaner way to do this is check if the element is hidden or not.That would more directly relate to what the code is trying to accomplish. I think you can check what the current style.display is set to.

portfolio/src/main/webapp/script.js
```diff
@@ -199,16 +244,69 @@ async function deleteMessage(msg) {
...
+    return
+  }
+  
+  // Put reply form under the post.
+  let replyNode = document.getElementById('reply-form');
+  var referenceNode = document.querySelector(`#${CSS.escape(postEleID)}`);
+  referenceNode.parentNode.insertBefore(replyNode, referenceNode.nextSibling);
+  replyNode.style.display = 'block';
+  
+  // Show submit button only if a user types,
```
martiKallas, line 268: I think you can 'grey' out the buttons somehow. It might be possible to make the text box a required field which would do the same check as you have below.

martiKallas (approved):
Just a few comments. This could have been probably 2 PRs. For example, some of the visual changes could have been decoupled from the functional changes to sending and getting replies. The changes and style look great though!

portfolio/src/main/webapp/script.js
Andrewmob2703-zz: I tried passing the form data itself but the server was not able to read the values (returning null). And i googled and saw that it's possible to pass the form data to a URLSearchParams, which worked. For time's sake i didn't try to find why form data wasn't working but i think I needed to specify some headers or probably use json.stringify 
I can try that later

portfolio/src/main/webapp/script.js
Andrewmob2703-zz: Sounds good. I'll try that later

portfolio/src/main/webapp/script.js
Andrewmob2703-zz: I think that way if a reply button is shown for some post and user clicks the reply button of another post, the reply box will be hidden.(because it's displayed somewhere), which is what was happening but I didn't like it. But in this case we want to show the box (no matter the box's display status) since the user is trying to reply to some other post. That's why the previous condition checks if the button clicked is for a different post. 
I don't know if that makes any difference.

portfolio/src/main/webapp/script.js
Andrewmob2703-zz: Oh yeah grey-ing the button out would be great

PR #59, opened 2020-07-31
Use different IDs but same names for both forms so server can use sam…

Andrewmob2703-zz:
…e constants to get data.
Also:
Remove nickname box from both forms, and 
Use form's own submit button to submit form as normal

### DIFF
--- a/portfolio/src/main/webapp/index.html
+++ b/portfolio/src/main/webapp/index.html
@@ -5,15 +5,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>BIPOC Blog</title>
     <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
- 
-    <style>
-      /* Form is hidden by default. */
-      .hidden {
-        display: none;
-      }
-    </style>
   </head>
+
   <body onload="fetchBlobstoreUrlAndShowForm(); defaultPage(); loadPosts();">
     <button class="tablink" onclick="openPage('Home', this, 'grey')" id="defaultOpen">Home</button>
     <button class="tablink" onclick="openPage('Blog', this, 'grey')">Blog</button>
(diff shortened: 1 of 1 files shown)


portfolio/src/main/webapp/index.html
```diff
@@ -5,15 +5,8 @@
...
     <title>BIPOC Blog</title>
     <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
- 
-    <style>
-      /* Form is hidden by default. */
-      .hidden {
-        display: none;
-      }
-    </style>
```
Andrewmob2703-zz, line 15: took this to css file

portfolio/src/main/webapp/index.html
```diff
@@ -65,31 +55,30 @@ <h1>Blog</h1>
...
-          <textarea id="text-input-reply" rows="6" cols="60" placeholder="What's on your mind..."></textarea>
+        <form id="reply-form" enctype="multipart/form-data">
+          <textarea id="reply-text-input" name="text-input" rows="3" cols="60" placeholder=""></textarea>
           <br>
-          <div id="hidden-reply">
-            <input type="text" id="parentID" value="0"><br>
-            <input type="text" id="tags-reply" value="#general"><br>
+          <div class="hidden">
+            <input type="text" id="reply-parentID" name="parentID" value="0"><br>
+            <input type="text" id="reply-tag" name="tags" value="#general"><br>
```
martiKallas, line 75: Just a thought - should this value be assigned as the Parent comment's tag? Does that already happen somewhere?

martiKallas (approved):
LGTM

portfolio/src/main/webapp/index.html
Andrewmob2703-zz: Yeah it happens inside the handleReplies function in script.js

PR #61, opened 2020-07-31
Make getUserNickname a public static 

Andrewmob2703-zz:
so that dataservlet can access this.
Also add license lines to this file


peizhaoou (commented):
Which class needs to access it?

Andrewmob2703-zz:
> Which class needs to access it?

DataServlet class

peizhaoou:
Added a related comment in your other PR. You should keep the two in the same PR and even the same commit for changes like this next time because you would want every commit to at least compile and work.

peizhaoou (approved):

Andrewmob2703-zz:
> Added a related comment in your other PR. You should keep the two in the same PR and even the same commit for changes like this next time because you would want every commit to at least compile and work.

Got it

PR #62, opened 2020-07-31
Update DataServlet to get user nickname from LoginServlet

Andrewmob2703-zz:
Also arrange doPost to make logic flow in order


peizhaoou (approved):
Now I know where we need to use LoginServlet.getUserNickname(). I think the other PR that updates the method to public static and this one should definitely be grouped together so that you can ensure the head of master always compile! It is fine this time, but I think we should definitely always make changes that ensure the head will compile and work after each merge.

PR #68, opened 2020-08-03
Add parameter constants

Andrewmob2703-zz:
These are used in the delete function in script and deleteServlet


martiKallas (approved):
Nice quick fix!

PR #69, opened 2020-08-03
Move putBlogsInDatastore to DatastoreUtils.java

Andrewmob2703-zz:


martiKallas (changes requested):
Can you write a test for this function? It should just be confirming that if you call this function, then datastore contains an entry with the correct parameters. See Tayyaba's tests and https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting#datastore-memcache

martiKallas (approved):

PR #81, opened 2020-08-05
Added test for doGetFromDatastore

smarquez-jpg:


portfolio/src/test/java/com/google/sps/DatastoreUtilsTest.java
```diff
@@ -188,4 +188,11 @@ public void PutBlogsInDatastoreTest_withoutPost() {
     assertEquals(0, ds.prepare(new Query(BlogConstants.BLOG_ENTITY_KIND)).countEntities(withLimit(5)));
   }
 
+  @Test
+  public void testDoGetFromDatastore() {
+    createDatastoreEntities();
+    LinkedList<BlogMessage> actual = DatastoreUtils.doGetFromDatastore(1);
+
+    Assert.assertEquals(1, actual.size());
```
martiKallas: You need some extra checks here. You want this to fail if doGetFromDatastore changes in an unwanted way. doGetFromDatastore has 3 functions that it calls: putRepliesWithPosts, updateTagsToSearch, and sortAndLoadfromBlogHashmap. You will need to check the output from doGetFromDatastore to confirm that each of these are called.

martiKallas (changes requested):
Thanks for starting this, Steven! Feel free to get help from your podmates too.

Andrewmob2703-zz (commented):
I probably should have tested this since I added new functionality.
New Tests Added.

martiKallas (approved):
Nice changes!

I still don't think this tests if replies get attached to posts, but feel free to submit in the interest of time.

Andrewmob2703-zz:
> Nice changes!
> 
> I still don't think this tests if replies get attached to posts, but feel free to submit in the interest of time.

I think Steven tested that already in lines 100-112.

PR #84, opened 2020-08-06
Add Event listener on post form, and

Andrewmob2703-zz:
remove styles from the functions, and
Add tag elements to the posts


peizhaoou (approved):
I only briefly looked at the change at a high level. Given the time constraint, I think you can submit as long as you have manually tested this.

PR #86, opened 2020-08-07
Remove image string from the HashTableTest.java

Andrewmob2703-zz:
Also make replies filed a BlogMessage instead of String


peizhaoou (approved):
Just a note: in reality, you would want to have this change together with the PR you updated the BlogMessage's constructor because you don't want any breaking changes to be checked in to the repo, including the tests.

